PR #7, opened 2024-05-22
Add Overlay component and stubbed FilterMenu component

### LINKED ISSUE #5: Geometry filter menu
Add the Filter Menu with support for all screen sizes. Mock ups available [here](https://www.figma.com/file/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?type=design&node-id=26-12737&mode=design&t=1C2YBnzdu9rXZIUU-0).

### Breakpoints and Responsive Styling
The designs are given in two screens sizes - one to show a portrait orientation mobile phone and one for "desktop" monitor sizes. For implementation, we'll use the `lg` breakpoint already present in our [design system repo](https://github.com/NYCPlanning/ae-streetscape/blob/main/src/theme/tokens/breakpoints.ts#L5). At `62em` (equivalent to `992px`), this will effectively allow us to apply one set of styles for screen at portrait tablet sizes and another for sizes larger than that.

### Sizing and padding
The margins and paddings as you see them in Figma aren't necessarily pixel-perfect. I'll put notes about specific margins, paddings, widths, and heights where applicable below. Where not noted, use your best judgement based on the mock ups and we'll make adjustments once a PR is up.

### Code Organization
I won't be too prescriptive about component design here but this should be a good candidate for some sort of `FilterMenu` component that accommodates all screen sizes. As a general rule, we want to avoid creating different components for specific screen sizes where possible.

## Desktop behavior

- [ ] Filter Menu is always visible
- [ ] "Filter by Geography" button is always hidden
- [ ] Filter Menu has fixed `21.25rem` width
- [ ] Filter Menu panel has 1rem (use `4` token from design system) padding on left and right
- [ ] Height should be determined by height of content (`auto` in css)

## Mobile behavior

- [ ] ~~Hide menu and show "Filter by Geography" button when user taps anywhere outside the menu~~ Cutting this for now in favor of simple clickable chevron icon to close menu
- [ ] ~~Hide menu and show "Filter by Geography" button when user swipes down on the menu~~ Cutting this for now in favor of simple clickable chevron icon to close menu
- [ ] Filter Menu panel has 0.75rem (use `3` token from design system) padding on left, and right sides, and `6` on top and bottom
- [ ] Filter Menu is horizontally centered and has `21.25rem` _max_ width.
- [ ] Height should be determined by height of content (`auto` in css)
- [ ] When visible the Filter Menu renders at the bottom of the `Overlay` component, similar to existing "Filter by Geography" button

## Form requirements

- [ ] Have three Selects - Geography Type, Borough, District Number
- [ ] Geography type should have "Community Districts" and "City Council Districts" options
- [ ] Borough and District Number have no options (will be added via future Issues once hooked up to API)
- [ ] Borough and District Number are disabled when no Geography Type is selected
- [ ] Has "Go to Selected Geography" button. Because this should be disabled until the form is populated and the user cannot fill out the form until we add boroughs and districts, this button can be just set to disabled for this Issue.

TylerMatteo:
## Summary
* Add `<meta>` tag in `root.tsx` to fix issue with scaling on smaller screen sizes
* Add new `<Overlay>` component that sits "on top" of the map and add universal padding for panel elements
* Stubs out new responsive `<FilterMenu>` component as well as the button to toggle showing it on mobile
* e2e tests for `<FilterMenu>`. I haven't added any unit tests for the component yet because I couldn't identify anything worth testing yet but open to feedback on this.

Closes #5


app/components/Overlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+          pointerEvents: "auto",
+        },
+      }}
+    >
+      <Show above="lg">
+        <FilterMenu onClose={() => setShouldShowFilterMenu(false)} />
+      </Show>
+      {shouldShowFilterMenu ? (
+        <Hide above="lg">
+          <FilterMenu onClose={() => setShouldShowFilterMenu(false)} />
```
TylerMatteo, line 29: I don't love rendering this component twice to get the responsive behavior but it was the only approach I could come up with that didn't require installing a 3rd party hook to get the screen size in pixels as state.

TangoYankee:
1) Clicking on the dropdown arrow appears to be a dead spot for clicks
2) The geography selector is covering the attribution
3) The mockups took a left turn to show "borough" and "district" when CDs are shown and "district" when ccds are shown

TangoYankee:
@TylerMatteo It's a small detail but helps with QOL from looking at tickets on the board that need review. Could you use the [`closes` syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) in the description to link the ticket and the PR?

TylerMatteo:
@TangoYankee I pushed implementing hiding the borough dropdown for CCDs. I'll create a separate issue in the Streetscape repo for the `<Select>` bug. For the attribution, I think we can table that for the purposes of this PR. I didn't hide it because I'd prefer not have it be unreachable in the mean time.

app/components/FilterMenu.tsx
```diff
@@ -0,0 +1,86 @@
...
+  Flex,
+  FormControl,
+  FormLabel,
+  Heading,
+  Show,
+  Hide,
+  Select,
+  HStack,
+} from "@nycplanning/streetscape";
+import { ChevronDownIcon } from "@chakra-ui/icons";
```
TangoYankee, line 12: Do we want these icons to be packaged in streetscape as well?

TangoYankee:
My major concerns have been addressed. 

Will you squash this into one commit or do you think it's better to keep as two?

TylerMatteo:
Squashed and pushed

TangoYankee (approved):

PR #25, opened 2024-06-24
Add getCursor handler to Atlas to show pointer cursor on pickable features

TylerMatteo:
This PR is a small change that adds a custom [getCursor](https://deck.gl/docs/api-reference/core/deck#getcursor) handler to the `<Deck>` instance in `Atlas`. I noticed that with #23, the cursor still showed up as "drag" even when the user was hovering over pickable features. I think this change provides feedback to the user that they can click on projects.

### DIFF
--- a/app/components/atlas.client.tsx
+++ b/app/components/atlas.client.tsx
@@ -29,6 +29,12 @@ export function Atlas() {
         communityDistrictsLayer,
         cityCouncilDistrictsLayer,
       ]}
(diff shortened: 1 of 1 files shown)


TangoYankee (approved):
I find it odd that if you're dragging while hovering then the hovering pointer takes precedence. But, I don't see an easy way to streamline this behavior. 

Overall, the cursor behavior in this PR is already an improvement over the existing behavior.

TangoYankee (approved):

Issue #12, opened 2024-05-31
Show Council Districts layer on map


TylerMatteo:
# Description
User sees Council District outlines on map when City Council District filter type is selected in Filter Menu.

## Acceptance Criteria

- [ ] When user selects "City Council Districts" from District Type dropdown in `FilterMenu` component, user sees CCD outlines and labels on the map
- [ ] District boundary styles match [specs in Figma](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=670-5575&t=bHnqp3D5d54Ghn2t-0)

## Out of scope
* Additional map styling for _selected_ CCD - to be handled in separate Issue for map filtering

## Related work
* Blocked by https://github.com/NYCPlanning/ae-zoning-api/issues/254

TangoYankee:
blocked by #18

TylerMatteo:
This should be unblocked now that the tiles endpoint is implemented and we have the changes in PR #27 merged into the front end. The changes in that PR hook up the `<FilterMenu>` component to the router such that the selected type is tracked by the `districtType` query parameter. Implementation for this Issue can use that query param to control the `visible` prop on a new `MVTLayer` for CCDs.

PR #27, opened 2024-06-25
Add separate _index page for mobile filtermenu layout and connect district type to query params

TylerMatteo:
## Summary

1. Adds new root `_index` route that renders a `<FilterMenu>` and the `Show Geography filter` button only on mobile. In combination with the `<FilterMenu>` instance in `root.tsx`, this is mean to accomplish the desired behavior where the filter menu is always shown on desktop sizes but is hidden on mobile on non-root routes (so that content panel can be shown).
2. Connected the district type dropdown to store it's state in query param.

This PR is intended to set up just enough of the FilterMenu component to unblock other works, primarily #12 and #13

### DIFF
--- a/app/components/FilterMenu.tsx
+++ b/app/components/FilterMenu.tsx
@@ -1,4 +1,5 @@
-import { useState, FormEvent } from "react";
+import { FormEvent } from "react";
+import { useSearchParams, useNavigate } from "@remix-run/react";
 import {
   Button,
   Flex,
@@ -12,10 +13,13 @@ import {
 } from "@nycplanning/streetscape";
 import { ChevronDownIcon } from "@chakra-ui/icons";
 
-export type GeographyType = "cd" | "ccd" | null;
-
-export const FilterMenu = ({ onClose }: FilterMenuProps) => {
-  const [geographyType, setGeorgaphyType] = useState<GeographyType>("cd");
+export const FilterMenu = ({
+  onClose = () => {
+    return;
+  },
+}: FilterMenuProps) => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const districtType = searchParams.get("districtType");
   return (
     <Flex
       borderRadius={"base"}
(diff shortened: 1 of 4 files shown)


app/components/FilterMenu.tsx
```diff
@@ -49,25 +49,41 @@ export const FilterMenu = ({ onClose }: FilterMenuProps) => {
...
-            value={geographyType}
+            value={districtType === null ? "" : districtType}
           >
             <option value={"cd"}>Community District</option>
             <option value={"ccd"}>City Council District</option>
           </Select>
         </FormControl>
         <HStack spacing={2} width={"full"}>
-          {geographyType !== "ccd" ? (
+          {districtType !== "ccd" ? (
```
TangoYankee, line 86: I'm familiar with the advice to use the full ternary operator in these cases to avoid the embarrassment of accidentally rendering "0". However, this is an actual boolean check. We can safely use the "&&" operator. This can help clean up the trailing "null" and keep things concise as we add more conditions.

app/components/Overlay.tsx
```diff
@@ -22,30 +24,13 @@ export const Overlay = () => {
       }}
     >
       <Show above="lg">
-        <FilterMenu onClose={() => setShouldShowFilterMenu(false)} />
+        <FilterMenu
+          onClose={() => {
+            console.log("close desktop");
```
TangoYankee, line 29: Should this be another placeholder function?

app/routes/_index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useState } from "react";
+
+export default function Index() {
+  const [shouldShowFilterMenu, setShouldShowFilterMenu] = useState(false);
+  return (
+    <Hide above="lg">
+      {shouldShowFilterMenu ? (
+        <FilterMenu
+          onClose={() => {
+            console.log("close mobile");
```
TangoYankee, line 12:
```suggestion
```

app/components/FilterMenu.tsx
```diff
@@ -12,10 +13,9 @@ import {
 } from "@nycplanning/streetscape";
 import { ChevronDownIcon } from "@chakra-ui/icons";
 
-export type GeographyType = "cd" | "ccd" | null;
-
 export const FilterMenu = ({ onClose }: FilterMenuProps) => {
-  const [geographyType, setGeorgaphyType] = useState<GeographyType>("cd");
+  const [searchParams, setSearchParams] = useSearchParams();
+  const districtType = searchParams.get("districtType");
```
TangoYankee, line 18: More precise typing

```suggestion
  const districtType = searchParams.get("districtType") as null | "ccd" | "cd";
```

TangoYankee (approved):
None of the comments are necessary changes

PR #41, opened 2024-07-08
Add capital project detail panel and connect it to route for single project

### LINKED ISSUE #16: Select project from map and show project details
## Acceptance Criteria
- [x] User can click or tap a project from the map and see the detail panel for that project ([Figma](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=337-3601&t=i9OdRPOzwaKcOnqa-0))
- [x] Project detail panel is styled according to Figma designs for all screen sizes.
- [ ] When user selects a projects, map is panned so that the selected project is centered
- [x] Styling of project on the map updates to reflect selected state ([Figma](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=670-5575&t=i9OdRPOzwaKcOnqa-0))

### Out of scope
- Showing [commitment-level data](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=337-3772&t=i9OdRPOzwaKcOnqa-0) for the project. To be covered in #17 

## Blocked by
- #8 
- #18

TylerMatteo:
Closes #16 

## Summary

This PR builds out the panel for showing project details for a single project. It uses the existing `/capital-projects/:managing-code/:capital-project-id` route. 

### New dependencies
* [Numbro](https://numbrojs.com/getting-started.html) - Used for formatting the future commitments total.
* [date-fns](https://date-fns.org/) - Used for date formatting and logic. If you've worked with Moment before, you can think of date-fns as serving a similar niche. I like it because it is function-oriented and based around the browser-native `Date` class.

## Not included
There is one requirement in #16 not included here - the "fly to" functionality where the map moves to a bounding box of the selected project. I'm going to capture the issues I had implementing that in a separate Discussion or Issue and put forward some potential solutions and next steps.


app/routes/capital-projects.$managingCode.$capitalProjectId.tsx
```diff
@@ -1,3 +1,41 @@
...
 export default function CapitalProject() {
-  return <></>;
+  const navigate = useNavigate();
+  const { capitalProject, agencies } = useLoaderData<typeof loader>();
+  return (
+    <CapitalProjectDetailPanel
+      capitalProject={capitalProject}
+      agencies={agencies}
+      onClose={() => {
+        navigate("/");
```
TangoYankee, line 37: This will clear any query parameters and reset the state of the geography menu. My understanding is that we're trying to decouple the content panel from the geography menu. To honor this, we should preserve the query parameters. The capital projects layer has an example of how to [pass the existing parameters to the new route](https://github.com/NYCPlanning/ae-cp-map/blob/62a9766562b40b5a31c99067f6ae1f082479b4f4/app/components/layers/useCapitalProjectsLayer.client.tsx#L48).

setup-test.ts
```diff
@@ -1,18 +1,31 @@
...
  */
-import { handlers } from "./app/gen";
-import { setupServer } from "msw/node";
+import { handlers } from './app/gen';
+import { setupServer } from 'msw/node';
 
 window.URL.createObjectURL = vi.fn();
+Object.defineProperty(window, 'matchMedia', {
+  writable: true,
+  value: vi.fn().mockImplementation((query) => ({
```
TangoYankee, line 11: If I'm reading this correctly, we do this to mock browser methods and silence deck errors. I was hesitant to try this with my branch when I ran into these errors because I have concerns about "chasing errors". Basically, deck will continue to evolve and we will have to keep updating the mocks to reflect the methods invoked by deck. But, I suppose we can try this approach now and revisit it only if "chasing errors" actually becomes a real problem.

app/routes/capital-projects.$managingCode.$capitalProjectId.tsx
```diff
@@ -1,3 +1,41 @@
...
+import { CapitalProjectDetailPanel } from "../components/CapitalProjectDetailPanel";
+
+export async function loader({ params }: LoaderFunctionArgs) {
+  const agenciesResponse = await findAgencies({
+    baseURL: `${import.meta.env.VITE_ZONING_API_URL}/api`,
+  });
+  if (
+    typeof params.managingCode === "undefined" ||
+    typeof params.capitalProjectId === "undefined"
+  ) {
```
TangoYankee, line 16: Checking the type is "undefined" strikes me as more verbose and less precise than directly checking the value is literally undefined. Also, we reference each value at least twice. We can de-structure it once and directly use the primitive values; ie)
```suggestion
const { managingCode, capitalProjectId } = params;
  if (
    managingCode === undefined ||
    capitalProjectId === undefined
  ) {
```

app/components/CapitalProjectDetailPanel/CapitalProjectDetailPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+  onClose,
+}: CapitalProjectDetailPanelProps) => {
+  const [isExpanded, setIsExpanded] = useState(false);
+  const getFiscalYearForDate = (date: Date): number => {
+    const year = getYear(date);
+    const month = getMonth(date);
+    return month <= 6 ? year : year + 1;
+  };
+
+  const formatFiscalYearRange = (minDate: Date, maxDate: Date) => {
```
TangoYankee, line 36: This appears to be a "pure function" (it has no reliance on the component state). We should be able to move it to outside of the component. This will avoid it being redefined on each re-render

app/components/CapitalProjectDetailPanel/CapitalProjectDetailPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+    if (compareAsc(minDate, maxDate) === 0) {
```
TangoYankee, line 37: This seems like fun opportunity for a ternary operator. But, that's purely a matter of preference

app/components/CapitalProjectDetailPanel/CapitalProjectDetailPanel.tsx
```diff
@@ -0,0 +1,181 @@
...
+  onClose: () => void;
+}
+
+export const CapitalProjectDetailPanel = ({
+  capitalProject,
+  agencies,
+  onClose,
+}: CapitalProjectDetailPanelProps) => {
+  const [isExpanded, setIsExpanded] = useState(false);
+  const getFiscalYearForDate = (date: Date): number => {
```
TangoYankee, line 30: Another pure function that can be moved outside the component

TangoYankee:
Unable to verify the behavior locally because I received the following error when running:

this style stylesheet is undefined

TylerMatteo:
> Unable to verify the behavior locally because I received the following error when running:
> this style stylesheet is undefined

This PR previously included an upgrade to `maplibre-gl` 4.5 from 4.4.1. 4.5 seems to have introduced a [bug](https://github.com/maplibre/maplibre-gl-js/issues/4340). I fixed it by pinning the dependency to `~4.4.1`.

setup-test.ts
TylerMatteo: Yeah I agree with the sentiment but this seems to be the best approach for now. Googling the error I was getting led me to [this Vitest Issue](https://github.com/vitest-dev/vitest/issues/821). It's probably fairer to say the problem is actually with `jsdom`, the DOM implementation Vitest uses under the hood. They haven't implemented every browser API, and this is the recommended way of filling missing ones as-needed.

TangoYankee:
The animation when expanding and shrinking the panel is a nice touch

app/components/CapitalProjectDetailPanel/CapitalProjectDetailPanel.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  it("should render the detail panel with project description", () => {
+    render(
+      <CapitalProjectDetailPanel
+        capitalProject={capitalProject}
+        agencies={agencies}
+        onClose={onClose}
+      />,
+    );
+    expect(screen.getByText(capitalProject.description)).toBeVisible();
```
TangoYankee, line 34: Hmm, I'm intrigued by asserting that it's visible and not just in the document

app/components/CapitalProjectDetailPanel/CapitalProjectDetailPanel.test.tsx
```diff
@@ -0,0 +1,84 @@
+import { CapitalProjectDetailPanel } from "./CapitalProjectDetailPanel";
+import {
+  CapitalProjectBudgeted,
+  createCapitalProjectBudgeted,
+  Agency,
+} from "~/gen";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+describe("CapitalProjectDetailPanel", () => {
```
TangoYankee, line 10: Interesting, we group tests differently. I'm lazy and group all the "thing should be there" assertions together in a single test. It seems you more strictly follow the "one assertion per test". Respect. I suppose I'm more lax about grouping them because I feel the assertions are expressive enough that I can identify which assertion failed within a test. Anyway, neat.

app/routes/capital-projects.$managingCode.$capitalProjectId.tsx
```diff
@@ -1,3 +1,43 @@
...
+import {
+  findCapitalProjectByManagingCodeCapitalProjectId,
+  findAgencies,
+} from "../gen";
+import { CapitalProjectDetailPanel } from "../components/CapitalProjectDetailPanel";
+
+export async function loader({ params }: LoaderFunctionArgs) {
+  const { managingCode, capitalProjectId } = params;
+  const agenciesResponse = await findAgencies({
+    baseURL: `${import.meta.env.VITE_ZONING_API_URL}/api`,
```
TangoYankee, line 12: There's gotta be a way to configure the baseUrl so that we don't need to state it on each call.

TangoYankee: Issue #46

app/routes/capital-projects.$managingCode.$capitalProjectId.tsx
```diff
@@ -1,3 +1,43 @@
...
+  });
+  if (managingCode === undefined || capitalProjectId === undefined) {
+    throw json("Bad Request", { status: 400 });
```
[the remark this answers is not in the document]
TangoYankee: #47

TangoYankee (approved):

Issue #36, opened 2024-07-02
Update `FilterMenu` component to new designs


TylerMatteo:
## Acceptance Criteria

- [ ] Update existing `FilterMenu` component to new designs ([mobile](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1329-23360&t=GiD5KJgOJ5K7t9Md-0) and [desktop](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1187-32851&m=dev))
- [ ] Use `Accordion` component from Streetscape to make filter menu collapesable
- [ ] Update logic for borough dropdown - Current logic is that it is hidden when City Council District district type is selected, new logic is that it remains but is disabled when CCD is selected
- [ ] Collapsable FilterMenu is always shown on mobile - note difference with old design that used "Filter by Geography" button. That button should be removed

### Out of scope
* "About" panel - on [desktop](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1187-32851&m=dev), the new designs have the "Capital Projects Portal" about info as another accordion section below the filter menu. For this Issue, only implement the filter menu part of that accordion. The about info portion will be added in #6 

### Block by
#28

dhochbaum-dcp:
Ready for QA: https://qa-capitalplanning.nycplanningdigital.com

TylerMatteo:
@wndyli @jessicashanshanhuang This is ready for feedback at that URL ☝ Please self assign and move to "In QA" once you've started testing.

jessicashanshanhuang:
Hey @dhochbaum-dcp! Here's our feedback:
<img src="https://github.com/user-attachments/assets/c1f9080b-5bd8-4620-be15-f39741d1181c" width=40% height=40%>

<img width="1344" alt="Screenshot 2024-07-31 at 5 10 29 PM" src="https://github.com/user-attachments/assets/177bfa95-7d37-4923-bbb7-c4914c8e074a">

In the City Council District Option, for the District Number dropdown:
- If it's possible to show only five options at a time and include a scrollbar to go through the selection, that would be more optimal to keep consistent dropdowns and not have that element expand over the webpage. 
[...]

PR #48, opened 2024-07-10
36 Update FilterMenu component to new designs

dhochbaum-dcp:
Need feedback on positioning on mobile, we currently have logo and mapcontrols overlapping where the filter is in the designs.

You may have a better way of setting the defaultIndex than duplicating the Accordion in a Show/Hide block - if you do, please let me know.  I attempted to use `useMediaQuery` to determine whether the page was mobile, as is done in atlas.client.tsx:31; unfortunately what seemed to occur was that `useMediaQuery("(max-width: 767px)")[0]` always returns `false` on the initial render, then updates to `true`.  Because of this, setting the property with `defaultIndex={useMediaQuery("(max-width: 767px)")[0] ? "this is mobile" : 0}` will always set the defaultIndex to 0.

- Update existing FilterMenu component to new designs ([mobile](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1329-23360&t=GiD5KJgOJ5K7t9Md-0) and [desktop](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1187-32851&m=dev))
- Use Accordion component from Streetscape to make filter menu collapesable
- Update logic for borough dropdown - Current logic is that it is hidden when City Council District district type is selected, new logic is that it remains but is disabled when CCD is selected
- Collapsable FilterMenu is always shown on mobile - note difference with old design that used "Filter by Geography" button. That button should be removed

Completes #36


TylerMatteo:
@dhochbaum-dcp it looks like you introduced a merge commit with [b4e1088](https://github.com/NYCPlanning/ae-cp-map/pull/48/commits/b4e1088d9652375d7f43997e633aabb521586a65). We want to stay away from that so that we get a linear history on `main` and don't have PRs where changes that aren't actually new show up in the diff. You'll have to rebase these changes onto main and perhaps do some squashing.

app/root.tsx
```diff
@@ -185,15 +184,32 @@ export default function App() {
...
-                />
+                <Hide above="lg">
+                  <FilterMenu>
+                    <AdminDropdowns />
+                  </FilterMenu>
+                </Hide>
+                <Flex
+                  direction={{ base: "column-reverse", lg: "column" }}
+                  justify={{ base: "flex-start", lg: "space-between" }}
+                  align={"flex-end"}
```
TylerMatteo, line 199: Adding `pointerEvents={"none"}` here should fix the issue where this Flex is blocking interaction with the map.

app/components/FilterMenu.tsx
```diff
@@ -1,59 +1,55 @@
...
+    background={"white"}
+    direction={"column"}
+    width={{ base: "full", lg: "21.25rem" }}
+    maxW={{ base: "21.25rem", lg: "unset" }}
+    boxShadow={"0px 8px 4px 0px rgba(0, 0, 0, 0.08)"}
+    defaultIndex={defaultIndex}
+  >
+    <AccordionItem
+      borderTopWidth="0"
+      borderBottomWidth="0 !important"
```
TylerMatteo, line 27: Let's avoid `!importants` if at all possible. What issues are introduced if you get rid of this?

TylerMatteo (changes requested):

app/components/FilterMenu.tsx
dhochbaum-dcp: image
Without it, the `:last-of-type` supersedes it.  Let me know if you'd prefer the extra pixel or the !important.

TylerMatteo (approved):

Issue #14, opened 2024-05-31
Filter and Show Projects list by Council District


TylerMatteo:
## Acceptance Criteria

- [x] When user selects a CCD from the Filter by District menu and clicks the "Go to Selected District" button, user is navigated to new `/city-council-districts/<cityCouncilDistrictId>/capital-projects` route and sees the content panel with the paginated list of projects
- [ ] ~~Projects on the map are filtered to only those that overlap with selected CCD. For projects that partially fall outside the CCD, all parts of the project should still be visible.~~ Moved to scope of #62 
- [ ] ~~Selected CCD boundary is highlighted ([figma](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=670-5575&t=KGRjmeAKGkxSYIRL-0))~~ Moved to scope of #62 
- [x] List of projects is paginated, with pagination arguments stored as query params. See `Page` [schema](https://github.com/NYCPlanning/ae-zoning-api/blob/main/openapi/openapi.yaml#L1120) from the underlying API for reference. Preserve other existing query params, such as those for keeping track of filter menu state.
- [x] User can click on a project to navigate to the nested route for that project: `/city-council-districts/<cityCouncilDistrictId>/capital-projects/<managingCode>/<capitalProjectId>`
- [ ] User can dismiss the project list detail panel to return to root `/` route

### Designs
[desktop](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1187-33249&t=GOfIYpfW8egMdy6G-0) and [mobile](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1329-23419&t=GOfIYpfW8egMdy6G-0)

### Out of scope
* Data export for projects within the CCD. Disregard "Export" button in figma designs for the purposes of this Issue.

TangoYankee:
Blocked by #18 and nycplanning/ae-zoning-api#251

TylerMatteo:
Partially blocked by #28, but work could begin on the detail panel by manually navigating to the `/city-council-districts/<cityCouncilDistrictId>` route.

TangoYankee:
@TylerMatteo 

Would it be more precise to make the route `/city-council-districts/<cityCouncilDistrictId>/capital-projects` and not just `/city-council-districts/<cityCouncilDistrictId>`? 

[...]

PR #56, opened 2024-07-24
14 filter capital projects by city council districts

pratishta:
Related to https://github.com/NYCPlanning/ae-cp-map/issues/14

### DIFF
--- a/app/components/CapitalProjectsList/CapitalProjectsAccordionPanel.tsx
+++ b/app/components/CapitalProjectsList/CapitalProjectsAccordionPanel.tsx
@@ -1,0 +1,69 @@
+import {
+  Accordion,
+  AccordionButton,
+  AccordionIcon,
+  AccordionItem,
+  AccordionPanel,
+  Box,
+  Flex,
+  Heading,
+} from "@nycplanning/streetscape";
+import { Agency, CapitalProject } from "~/gen";
+import { CapitalProjectsList } from "./CapitalProjectsList";
+
+export interface CapitalProjectsAccordionPanelProps {
+  capitalProjects: Array<CapitalProject>;
+  district: string;
+  agencies: Agency[];
+  children: React.ReactNode;
+}
+
+export const CapitalProjectsAccordionPanel = ({
+  capitalProjects,
+  district,
+  agencies,
+  children,
+}: CapitalProjectsAccordionPanelProps) => {
+  return (
+    <Flex
+      borderRadius={"base"}
+      padding={{ base: 3, lg: 4 }}
+      background={"white"}
+      direction={"column"}
+      width={{ base: "full", lg: "21.25rem" }}
+      maxW={{ base: "21.25rem", lg: "unset" }}
+      boxShadow={"0px 8px 4px 0px rgba(0, 0, 0, 0.08)"}
+      gap={4}
+    >
+      <Accordion defaultIndex={[0]} allowToggle>
+        <AccordionItem border="none">
+          <AccordionButton padding="0px" aria-label="Toggle project list panel">
+            <Box as="span" flex="1" textAlign="left">
+              <Heading
+                color="gray.600"
+                fontWeight={"bold"}
+                fontSize={"lg"}
+                paddingBottom={"8px"}
+              >
+                {district}
+              </Heading>
+            </Box>
+            <AccordionIcon size="lg" />
+          </AccordionButton>
+          <Box
+            borderTopWidth={"1px"}
+            borderTopColor={"gray.400"}
+            paddingBottom={4}
+          />
+          <AccordionPanel padding={"0px"}>
+            <CapitalProjectsList
+              capitalProjects={capitalProjects}
+              agencies={agencies}
+            />
+            {children}
+          </AccordionPanel>
+        </AccordionItem>
+      </Accordion>
+    </Flex>
+  );
+};
--- a/app/components/CapitalProjectsList/CapitalProjectsDrawer.tsx
+++ b/app/components/CapitalProjectsList/CapitalProjectsDrawer.tsx
@@ -1,0 +1,66 @@
+import { Box, Flex, Heading } from "@nycplanning/streetscape";
+import { Slide } from "@chakra-ui/transition";
+import { Agency, CapitalProject } from "~/gen";
+import { CapitalProjectsList } from "./CapitalProjectsList";
+import { useState } from "react";
+
+export interface CapitalProjectsDrawerProps {
+  capitalProjects: Array<CapitalProject>;
+  district: string;
+  agencies: Agency[];
+  children: React.ReactNode;
+}
+
+export const CapitalProjectsDrawer = ({
+  capitalProjects,
+  district,
+  agencies,
+  children,
+}: CapitalProjectsDrawerProps) => {
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  return (
+    <Slide direction="bottom" in={true} style={{ zIndex: 10 }}>
+      <Flex
+        padding={{ base: 3, lg: 4 }}
+        background={"white"}
+        direction={"column"}
+        borderTopLeftRadius={"base"}
+        borderTopRightRadius={"base"}
+        gap={4}
+      >
+        <Box
+          height={"4px"}
+          width={20}
+          backgroundColor={"gray.300"}
+          alignSelf={"center"}
+          role="button"
+          borderRadius={"2px"}
+          aria-label={
+            isExpanded
+              ? "Collapse project list panel"
+              : "Expand project list panel"
+          }
+          onClick={() => {
+            setIsExpanded(!isExpanded);
+          }}
+        />
+        <Heading color="gray.600" fontWeight={"bold"} fontSize={"lg"}>
+          {district}
+        </Heading>
+
+        <Flex
+          height={{ base: isExpanded ? "85vh" : "40vh", lg: "auto" }}
+          direction={"column"}
+          transition={"height 0.5s ease-in-out"}
+        >
+          <CapitalProjectsList
+            capitalProjects={capitalProjects}
+            agencies={agencies}
+          />
+          {children}
+        </Flex>
+      </Flex>
+    </Slide>
+  );
+};
--- a/app/components/CapitalProjectsList/CapitalProjectsList.tsx
+++ b/app/components/CapitalProjectsList/CapitalProjectsList.tsx
@@ -1,0 +1,55 @@
+import { Box, Flex, Text, VStack } from "@nycplanning/streetscape";
+import { useState } from "react";
+import { Agency, CapitalProject, CapitalProjectBudgeted } from "~/gen";
+import { CapitalProjectsListItem } from "./CapitalProjectsListItem";
+import { formatFiscalYearRange, currentDate } from "../../utils/utils";
+
+export interface CapitalProjectsListProps {
+  capitalProjects: CapitalProject[];
+  agencies: Agency[];
+}
+
+export const CapitalProjectsList = ({
+  capitalProjects,
+  agencies,
+}: CapitalProjectsListProps) => {
+  const listBody =
+    capitalProjects.length === 0 ? (
+      <Text>Reached end of projects</Text>
+    ) : (
+      capitalProjects.map((capitalProject) => {
+        return (
+          <CapitalProjectsListItem
+            key={`${capitalProject.managingCode}${capitalProject.id}`}
+            capitalProject={capitalProject as CapitalProjectBudgeted}
+            agency={
+              agencies.find(
+                (agency) => agency.initials === capitalProject.managingAgency,
+              )?.name
+            }
+            yearRange={formatFiscalYearRange(
+              new Date(capitalProject.minDate),
+              new Date(capitalProject.maxDate),
+            )}
+          />
+        );
+      })
+    );
+
+  return (
+    <>
+      <Box paddingBottom={4}>
+        <Text as={"span"}>
+          Mapped Capital Projects as of <Text as={"b"}>{currentDate()}</Text>
+        </Text>
+      </Box>
+      <Flex direction={"column"} overflow={"hidden"}>
+        <Box height={"70vh"} overflowY={{ base: "scroll" }}>
+          <VStack align={"start"} gap={3}>
+            {listBody}
+          </VStack>
+        </Box>
+      </Flex>
+    </>
+  );
+};
--- a/app/components/CapitalProjectsList/CapitalProjectsListItem.tsx
+++ b/app/components/CapitalProjectsList/CapitalProjectsListItem.tsx
@@ -1,0 +1,65 @@
+import {
+  Box,
+  HStack,
+  Heading,
+  VStack,
+  Text,
+  Flex,
+} from "@nycplanning/streetscape";
+import { Link, useLocation, useSearchParams } from "@remix-run/react";
+import { CapitalProjectBudgeted } from "~/gen";
+
+export interface CapitalProjectsListItemProps {
+  capitalProject: CapitalProjectBudgeted;
+  agency: string | undefined;
+  yearRange: string;
+}
+
+export const CapitalProjectsListItem = ({
+  capitalProject,
+  agency,
+  yearRange,
+}: CapitalProjectsListItemProps) => {
+  const [searchParams] = useSearchParams();
+  const location = useLocation().pathname;
+  return (
+    <Box w={"full"}>
+      <Link
+        to={{
+          search: `?${searchParams.toString()}`,
+          pathname:
+            location + `/${capitalProject.managingCode}/${capitalProject.id}`,
+        }}
+      >
+        <Flex
+          height={{ lg: "auto" }}
+          direction={"column"}
+          transition={"height 0.5s ease-in-out"}
+          gap={4}
+        >
+          <Box
+            backgroundColor="gray.50"
+            paddingY={3}
+            paddingX={2}
+            borderRadius={"base"}
+            width={"100%"}
+          >
+            <HStack align={"start"}>
+              <Flex flex={3}>
+                <VStack align="start">
+                  <Heading color="gray.600" fontWeight={"medium"}>
+                    {capitalProject.description}
+                  </Heading>
+                  <Text textStyle="micro">{agency}</Text>
+                </VStack>
+              </Flex>
+              <Flex flex={1}>
+                <Text textStyle="micro">{yearRange}</Text>
+              </Flex>
+            </HStack>
+          </Box>
+        </Flex>
+      </Link>
+    </Box>
+  );
+};
--- a/app/components/FilterMenu.tsx
+++ b/app/components/FilterMenu.tsx
@@ -11,7 +11,6 @@ import {
 export const FilterMenu = ({ children, defaultIndex }: FilterMenuProps) => (
   <Accordion
     allowToggle
-    allowMultiple
     borderRadius={"base"}
     padding={{ base: 3, lg: 4 }}
     background={"white"}
--- a/app/components/Pagination.tsx
+++ b/app/components/Pagination.tsx
@@ -1,0 +1,69 @@
+import { ChevronLeftIcon, ChevronRightIcon } from "@chakra-ui/icons";
+import { Box, HStack } from "@nycplanning/streetscape";
+import { Link, useSearchParams } from "@remix-run/react";
+import { setNewSearchParamsString } from "~/utils/utils";
+
+export interface PaginationProps {
+  total: number;
+}
+
+export const Pagination = ({ total }: PaginationProps) => {
+  const [searchParams] = useSearchParams();
+  const itemsPerPage = 7;
+  const pageParam = searchParams.get("page");
+  const page = pageParam === null ? 1 : parseInt(pageParam);
+  const canSkipBackward = page > 1;
+  const canSkipForward = total === itemsPerPage;
+
+  return (
+    <HStack gap={2}>
+      <Link
+        to={{
+          search: setNewSearchParamsString(searchParams, {
+            page: page - 1,
+          }),
+        }}
+      >
+        <Box
+          as="button"
+          disabled={!canSkipBackward}
+          _disabled={{ color: "grey" }}
+          fontSize={"xl"}
+          aria-label="left"
+        >
+          <ChevronLeftIcon />
+        </Box>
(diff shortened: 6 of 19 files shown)


app/routes/city-council-districts.$cityCouncilDistrictId.capital-projects.tsx
```diff
@@ -0,0 +1,78 @@
...
+      {
+        limit: 7,
+        offset,
+      },
+      {
+        baseURL: `${import.meta.env.VITE_ZONING_API_URL}/api`,
+      },
+    );
+
+  return {projectsByCityCouncilDistrictResponse, agencies: agenciesResponse.agencies, limit, offset, path, cityCouncilDistrictId};
```
TangoYankee, line 46: I suggest focusing the return values of the loader to server-specific data. This would translate to removing the "limit", "offset", "path", and "cityCouncilDistrictId" from here. If we need them again in the component, we can still access them through the component-specific hooks like "useSearchParams". 

The immediate practical benefit would be to clarify which values come from where. The API response is paginated and will have its own "limit" and "offset" parameters. Removing the client's "limit" and "offset" from the loader response may help clarify that they're coming from the client url and not the API response.

app/routes/city-council-districts.$cityCouncilDistrictId.capital-projects.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export default function CapitalProjectsByCityCouncilDistrict() {
+    const {projectsByCityCouncilDistrictResponse, agencies, cityCouncilDistrictId, limit, path, offset} = useLoaderData<typeof loader>();
+
+    return (
+        <>
+        <Show above='sm'>
+            <CapitalProjectsAccordionPanel
+                capitalProjects={projectsByCityCouncilDistrictResponse.capitalProjects}
+                limit={limit}
```
TangoYankee, line 57: I do like that we pass the limit and offset to the child components, rather than using the hooks within the children to access these values. In my opinion, only the `route` level components should be aware that they're on a specific path. Every child component should function independently of the route they're on. Passing the limit and offset parameters facilitates this.

app/components/Pagination.tsx
```diff
@@ -0,0 +1,67 @@
...
+    path: string,
+};
+
+export const Pagination = ({total, path}: PaginationProps) => {
+    // where do we want to limit to only having 7 as limit
+    const [searchParams, setSearchParams] = useSearchParams();
+    const limit = 7; // Number(searchParams.get("limit"));
+    const offset = Number(searchParams.get("offset"));
+
+    const currentPage = offset === 0 ? 1 : offset / limit + 1;
```
TangoYankee, line 19: Would this also give `1` in the case where offset is `0`? If so, it seems more succinct to me.

```suggestion
    const currentPage = offset / limit + 1;
```

app/components/Pagination.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const canSkipBackward = offset > 0;
+    const canSkipForward = total === limit;
+
+    // from https://www.jacobparis.com/content/remix-pagination
+    function setSearchParamsString(
```
TangoYankee, line 24: It seems this is a pure function (it relies solely on the parameters passed into it). If so, we can move its definition to outside the component. This will prevent it being redefined on every component render.

TangoYankee: Looking at this again, it seems the functionality as a whole is customizes the "useSearchParams" hook to update search parameters. I wonder if it is worth pulling out into a custom hook.

app/components/CapitalProjectsList/CapitalProjectsAccordionPanel.tsx
```diff
@@ -0,0 +1,58 @@
+import { Accordion, AccordionButton, AccordionIcon, AccordionItem, AccordionPanel, Box, Flex, HStack, Heading, IconButton, Stack } from "@nycplanning/streetscape"
+import { Agency, CapitalProject } from "~/gen";
+import { CapitalProjectsList } from "./CapitalProjectsList";
+
+export interface CapitalProjectsAccordionPanelProps {
+    capitalProjects: Array<CapitalProject>;
+    district: string;
+    limit: number;
```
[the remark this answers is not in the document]
pratishta: Nope this component was originally a superset of the current existing components and I forgot to pare down the params. I think for the most part these components don't need to know limit or offset. Only the Pagination component uses it and even then we can get them using the hook. Given the earlier comment about passing to children I'm not sure which way is better, drilling to Pagination or using the hook in that component only.

app/components/Pagination.tsx
pratishta: Yes exactly. I left it here as is to get it going but tbh, 1. I don't fully understand what's going on in the function (specifically just the `Record` type) and 2. I was wondering if we could customize this function to be a bit more specific to our needs or leave as is.

app/components/CapitalProjectsList/CapitalProjectsAccordionPanel.tsx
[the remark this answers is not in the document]
TangoYankee: Either method is valid. As much as prop drilling has become a 4 letter word, there's not actually a hard rule on when it goes from "the obvious way to pass props from a parent to a child" to "a tedious way to pass props from an ancestor to a descendant"

If we want to pass the props through the tree, we can save ourselves some tedium by using a spread operator. I have an example of this on the "Selected Button" components. The "goToSelected" function is passed using a spread operator on "props". [example spread operator](https://github.com/NYCPlanning/ae-cp-map/pull/57/files#diff-3ea55054f1e9839d3182faf506e063d23b5b2d0c35109f12b8fa81c8819b4192R10)

Another way to pass properties through several layers is to use context. There is a react [`useContext` hook](https://react.dev/learn/passing-data-deeply-with-context) and then [remix has a convenience wrapper over it](https://remix.run/docs/en/main/hooks/use-outlet-context), too. Though, is probably too cumbersome of a tool for this situation. It also makes the "Pagination" component dependent on being within a context. This is basically the same condition as the `useSearchParams` hook, where the component needs to exist within a route.

Finally, we can define the instance of the Pagination component inside the route component and pass the parameters directly from the route component to the Pagination component. The path parameters will then be enclosed within the pagination component. We then pass the component with its enclosed parameters as a child of the panel. [An example of this technique](https://github.com/NYCPlanning/ae-cp-map/pull/57/files#diff-3c42a24a48ac728b117dd702abb6dd3571ecd488fed0b7042b4f0997493efe24R76)

My personal preference is 1) Define pagination in the route 2) Use the spread operator to send the props through the tree 3) `useSearchParams` and other route hooks 4) `useContext`

pratishta:
This PR is ready for another round of reviews. Several things to note:

- I've left the commits as they are so I can check them out as needed. I'll squash when approved and before merging
- Testing needs some direction. I only added labels to elements specific to the components I was testing so that I could test them. 
- I'm not sure about this item from the issue "User can click on a project to navigate to the nested route for that project: /city-council-districts/<cityCouncilDistrictId>/capital-projects/<managingCode>/<capitalProjectId>". My impression is that each route path is specific to what it serves so I'm wondering if we need to create a new route file `city-council-districts.$cityCouncilDistrictId.capital-projects.$managingCode.$capitalProjectId` and have it render the individual project component...
- For this item in the issue "User can dismiss the project list detail panel to return to root / route", is this meant to happen when the user clicks on the up chevron? Does that make the panel disappear entirely? Or just collapsed? How does the user dismiss the drawer? With swiping the bar down? 
- I made it so that when the number of projects are a multiple of 7 and thus causing us to page to an empty list, we display a message that says "Reached the end of projects" I'm wondering if we need some designer guidance on that messaging

TylerMatteo:
@pratishta 

> I'm not sure about this item from the issue "User can click on a project to navigate to the nested route for that project: /city-council-districts//capital-projects//". 

Yep we do need a new route file for this that would render the individual route component. I'm hoping this is simple enough to include in the scope of this PR but open to breaking it out if needed. The correct name for that new file should be `city-council-districts.$cityCouncilDistrictId.capital-projects_.$managingCode.$capitalProjectId`. Notice the `_` in `capital-projects_`. That trailing underscore is how we tell Remix that we want to [nest the route path _without nesting the layouts of the routes_](https://remix.run/docs/en/main/file-conventions/routes#nested-urls-without-layout-nesting). From there, the contents of that new route should be more or less the same as what's in `capital-projects.$managingCode.$capitalProjectId.tsx`. You might even be able to re-export the `loader` function from that file.

pratishta:
@TylerMatteo 

Updated with the new route.

app/components/CapitalProjectsList/CapitalProjectsDrawer.tsx
```diff
@@ -0,0 +1,89 @@
+import { Box, Flex, Heading } from "@nycplanning/streetscape";
+import {
+  Drawer,
```
TylerMatteo, line 3: If you upgrade the installed Streetscape version to `^0.10.4`, you can get these component from there instead of importing from Chakra directly.

app/components/CapitalProjectsList/CapitalProjectsListItem.tsx
```diff
@@ -0,0 +1,44 @@
+import { Box, HStack, Heading, VStack, Text } from "@nycplanning/streetscape";
+import { Flex } from "@chakra-ui/react";
```
TylerMatteo, line 2: Import this from Streetscape instead.

package.json
```diff
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "@chakra-ui/icons": "^2.1.1",
+    "@chakra-ui/react": "^2.8.2",
```
TylerMatteo, line 21: Once you address other comments, you should be able to remove this and just rely on `streetscape`.

package.json
```diff
@@ -39,6 +40,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-map-gl": "^7.1.7",
+    "react-paginate": "^8.2.0",
```
TylerMatteo, line 43: Are you still using this library?

app/utils/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+      newSearchParams.delete(key);
+      continue;
+    }
+    newSearchParams.set(key, String(value));
+  }
+  // Print string manually to avoid over-encoding the URL
+  // Browsers are ok with $ nowadays
+  // optional: return newSearchParams.toString()
+  return newSearchParams.toString();
+  // return Array.from(newSearchParams.entries())
```
TylerMatteo, line 36: Making a note to clean up this commented code before approval. If you do go with just `return newSearchParams.toString();`, you can probably update that comment on line 33 as well?

app/components/FilterMenu.tsx
```diff
@@ -56,4 +57,5 @@ export const FilterMenu = ({
 export interface FilterMenuProps {
   onClose?: () => void;
   children: ReactNode;
+  toNav?: () => void;
```
TylerMatteo, line 60: Maybe there's a more generic name for this prop? Like `handleSubmit` or `onSubmit`? The component doesn't necessary "care" that we're using it to trigger navigation.

TylerMatteo (changes requested):

TangoYankee (commented):
This is a fun twist. The method we're using to keep query parameters does not distinguish between filter menu parameters and list parameters. So, when we navigate away from the list, we keep the `page=2` parameter even though we have left that context and it no longer makes sense to have that parameter

TangoYankee:
For the "As of date" we are showing the date of the current day. I am trying to find in the designs where we say what the date should be and I cannot. However,  I think it should be a date associated with the publishing of the data.

The current date isn't really a reflection of when the data was generated. In fact, I think it may cause confusion during periods where a new capital plan has been created but we haven't yet published it on the map. The date is implying we're showing data from today, when in reality we might be showing 4 month old data and it could even be an outdated version.

If we want "Current as of" data to come from the dataset,  we may want to add back the "ccp version" to the database and display that, rather than a specific date.

@TylerMatteo and @pratishta

pratishta:
@TangoYankee 

> The current date isn't really a reflection of when the data was generated. In fact, I think it may cause confusion during periods where a new capital plan has been created but we haven't yet published it on the map.

I agree. I used the current date as a placeholder to follow the designs. My inclination right now is to omit this date / bit of text all together until we know how we can get this date. Oor, it could be part of the hardcoded changes we make whenever we get a data update. Oooor if there's a way to set an env variable with whenever GHA runs the ETL

pratishta:
Ready for another round of reviews: 
* I used `Slide` instead of `Drawer` because currently there's no way to interact with the overlay (i.e. use the filter, move the map) with the Drawer open. Using slide transition was [recommended instead](https://github.com/chakra-ui/chakra-ui/issues/4081) 
* In mobile, when the slide is at full height, there is a bit of the filter menu peeking out. From the designs, we might want the filter menu to be a little lower on the view than it currently is, so that when the slide is at full height, it completely encompasses the filter menu. 
* On desktop, the accordion panel falls over the zoom in/out and compass widgets. Not sure how to prevent this overlapping. 
* Remaining notes from an earlier [comment](https://github.com/NYCPlanning/ae-cp-map/pull/56#issuecomment-2263220890)

TangoYankee:
> Ready for another round of reviews:
> 
>     * I used `Slide` instead of `Drawer` because currently there's no way to interact with the overlay (i.e. use the filter, move the map) with the Drawer open. Using slide transition was [recommended instead](https://github.com/chakra-ui/chakra-ui/issues/4081)
> 

This catches my eye as something I should probably integrate into the capital project detail panel and capital commitments work :thinking:

app/routes/city-council-districts.$cityCouncilDistrictId.capital-projects.tsx
```diff
@@ -1,3 +1,80 @@
...
+  const { projects, agencies, cityCouncilDistrictId } =
+    useLoaderData<typeof loader>();
+
+  const pagination = (
+    <Flex paddingTop={4} alignItems="center" justifyContent={"space-between"}>
+      <Pagination total={projects.total} />
+    </Flex>
+  );
+
+  const desktop = (
```
TylerMatteo, line 54: This is subjective but maybe the contents of this and `mobile` should just be in-lined inside the `return`? I tend to reserve creating these intermediate JSX variables to things that are rendered more than once (such as is the case here with `pagination`.

app/components/CapitalProjectsList/CapitalProjectsList.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const [isExpanded] = useState(false);
+  const [searchParams] = useSearchParams();
+  const location = useLocation().pathname;
+  const listBody =
+    capitalProjects.length === 0 ? (
+      <Text>Reached end of projects</Text>
+    ) : (
+      capitalProjects.map((capitalProject) => {
+        return (
+          <Box
```
TylerMatteo, line 26: Could this `Box` and `Link` be moved into `CapitalProjectsListItem`? It feels like that component would be more self-contained if it included those.

app/components/CapitalProjectsList/CapitalProjectsAccordionPanel.tsx
```diff
@@ -0,0 +1,74 @@
...
+      background={"white"}
+      direction={"column"}
+      width={{ base: "full", lg: "21.25rem" }}
+      maxW={{ base: "21.25rem", lg: "unset" }}
+      boxShadow={"0px 8px 4px 0px rgba(0, 0, 0, 0.08)"}
+      gap={4}
+    >
+      <Accordion defaultIndex={[0]} allowToggle>
+        <AccordionItem border="none">
+          <h2>
```
TylerMatteo, line 40: Is there a reason this uses `<h2>` instead of `<Heading>`? I've had issues with that component too but it is meant to replace using the native h1-6 elements using `as=`

TylerMatteo (changes requested):

app/components/CapitalProjectsList/CapitalProjectsList.tsx
pratishta: At the time I only wanted to pass in the strings that were going to be rendered in `CapitalProjectsListItem` (instead of passing id and managingCode and also calling the searchParams and location hooks) so that the list can handle the logic of assigning links

app/components/CapitalProjectsList/CapitalProjectsAccordionPanel.tsx
pratishta: Nope, I think it was a vestige of initial work, and far as I can tell, we don't even need it. Removing it

app/components/CapitalProjectsList/CapitalProjectsList.tsx
TylerMatteo: Gotcha. Can you move that into the list item component? I don't think it's an issue to have that component depend on those hooks if it needs them to do its job. You could have that component take in a `capitalProject` (of type `CapitalProjectBudgeted`), instead of passing in strings piecemeal.

TylerMatteo:
@pratishta last thing for me is just refactoring `CapitalProjectListItem` then lgtm.

app/components/CapitalProjectsList/CapitalProjectsDrawer.tsx
```diff
@@ -0,0 +1,66 @@
+import { Box, Flex, Heading } from "@nycplanning/streetscape";
+import { Slide } from "@chakra-ui/react";
```
TangoYankee, line 2: How are we using `Slide`?

I don't see us import `@chakra-ui/react` in the package.json

pratishta: Hm, I'm not really sure tbh. I wonder if it's because I never did an install after removing `@chakra-ui/react` from earlier feedback and my local modules remained the same. (I don't think it's coming from streetscape for what it's worth) 

I included `@chakra-ui/transition` now in`package.json`

TylerMatteo: This is likely because I had you remove the dependency in [this comment](https://github.com/NYCPlanning/ae-cp-map/pull/56#discussion_r1709857646) but now you _are_ depending on it again. I'm assuming it still works on your machine because `@chakra-ui/react` is still in your `node_modules`. Adding `@chakra-ui/transition` as a dependency is a good solution for now until we add `Slide` to Streetscape.

pratishta: Nice, just what I had suspected. 

Is there a typical way I could update my local node modules? A reinstall doesn't seem to get rid of ones that are no longer in `package.json`

horatiorosa: I think to get rid of unwanted modules, you probably would need to delete all the modules ( `rm -rf node_modules` ) and  reinstall them  to get a fresh, clean set of modules.

pratishta: Ooo didn't know about `npm prune` ty

TangoYankee: I opened to NYCPlanning/ae-streetscape#71 so that we can get the `slide` component from streetscape instead of Chakra. Though, I think we should wait to do said refactor until #71.

@pratishta and @TylerMatteo

pratishta: Thanks for that quick work. And yeah prefer to do refactoring of that nature not-in-this-ticket.

app/components/CapitalProjectsList/CapitalProjectsList.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <>
+      <Box paddingBottom={4}>
+        <Text as={"span"}>
+          Mapped Capital Projects as of <Text as={"b"}>{currentDate()}</Text>
+        </Text>
+      </Box>
+      <Flex direction={"column"} overflow={"hidden"}>
+        <Box
+          height={{ base: isExpanded ? "70vh" : "70vh" }}
```
TylerMatteo, line 50: Can you change this to `height={"70vh"}`

app/components/CapitalProjectsList/CapitalProjectsDrawer.tsx
```diff
@@ -0,0 +1,66 @@
...
+          onClick={() => {
+            setIsExpanded(!isExpanded);
+          }}
+        />
+        <Heading color="gray.600" fontWeight={"bold"} fontSize={"lg"}>
+          {district}
+        </Heading>
+
+        <Flex
+          height={{ base: isExpanded ? "85vh" : "40vh", lg: "auto" }}
```
TylerMatteo, line 53: There's a minor issue where the Pagination component shifts up on mobile when you expand the drawer. Changing `85vh` to `70vh` here seems to fix it but not sure if that breaks anything else. If you're reasonably confident that isn't going to cause any downstream effects, I think it's a good change to make now.

TylerMatteo (changes requested):

app/components/CapitalProjectsList/CapitalProjectsDrawer.tsx
pratishta: I don't think we should rely on setting a different height value to fix this. We should fix the placement of the filter menu first and then have a standardized height for the expansion before setting that value. See below for what I meant with the filter menu peeking out from my earlier comment. 
image

If we continue to render the pagination footer as a child we pass to the drawer or accordion component, we could probably have the Flexbox determine the anchoring of pagination with something like "align-content: space-between" or something. But I strongly feel that this work should be part of a separate issue.

TylerMatteo: Understood. As a solution for now that doesn't mess with the `vh` value, how about putting `marginTop={"auto"}` on the `pagination` Flex on line 49 in  city-council-districts.$cityCouncilDistrictId.capital-projects.tsx? That will make it anchor the the bottom of its parent.

pratishta: That works, I've updated 👍

TangoYankee: Hello 👋 , I noticed this behavior when integrating this project list component with the project detail panel. I believe I have a working solution. I'm writing some tests now and should have it ready for review later today.

TylerMatteo (approved):

Issue #62, opened 2024-08-07
Filter projects on map by City Council District


TylerMatteo:

TangoYankee:
Given the clipping issues with the "Mask" approach, we will likely want to pivot to the API filtering the tiles for us. With the api-based approach this ticket would depend on NYCPlanning/ae-zoning-api#365

PR #66, opened 2024-08-13
62 filter projects on map by ccd

pratishta:
Closes https://github.com/NYCPlanning/ae-cp-map/issues/62

Waiting for the ✅ from https://github.com/NYCPlanning/ae-cp-map/pull/65 to mitigate merge conflicts. Things to note in this PR:
- I pass an empty string to data instead of using the fallback geometry
- I also added the `hasCityCouncilDistrict` boolean to both capital project and ccd layer because I wanted to more explicitly tie the highlighting of the district and the filtered project tiles together in the code

### DIFF
--- a/app/components/atlas.client.tsx
+++ b/app/components/atlas.client.tsx
@@ -10,6 +10,7 @@ import {
   useCommunityDistrictsLayer,
   useCityCouncilDistrictsLayer,
   useCommunityDistrictLayer,
+  useCityCouncilDistrictLayer,
 } from "./layers";
 import type { MapView, MapViewState } from "@deck.gl/core";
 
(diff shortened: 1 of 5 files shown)


TangoYankee (changes requested):
This error occurs when using an empty string as the `data` value on a GeoJSONLayer. 

Screenshot 2024-08-13 at 12 38 39 PM


I know the `fallbackGeometry` feels hacky. But unless there is a significant issue with it, we should stick with it for now. It addresses the known error without causing knock-on problems. This frees our time to focus on the features that still need to be completed by the end of this sprint.

pratishta:
@TangoYankee Okay thanks for flagging what the error was. There's no significant issue with `fallbackGeometry` but if it's possible to use an empty value for a layer data at this point in dev, I think we should.  I used `[]` instead of `""` and I didn't get that error. Is that the case for you as well? (Looking at the default value here: https://deck.gl/docs/api-reference/core/layer#data)

TangoYankee:
> @TangoYankee Okay thanks for flagging what the error was. There's no significant issue with `fallbackGeometry` but if it's possible to use an empty value for a layer data at this point in dev, I think we should. I used `[]` instead of `""` and I didn't get that error. Is that the case for you as well? (Looking at the default value here: https://deck.gl/docs/api-reference/core/layer#data)

Thank you for pointing this out. In which case, yes it does make more sense to use an empty array than the fallback geometry. I'll update #65 to use the empty array instead of the fallback geometry

TangoYankee (approved):
:rocket:

PR #68, opened 2024-08-13
Implement flyto functionality

horatiorosa:
* Adds new `useCapitalProjectBudgetedGeoJsonLayer` hook for geojson layer displaying selected Capital Project
* Implements "fly to" functionality for council districts, community districts, and selected project with new `FlyToGeoJsonExtension`. When this extension is applied to a GeoJsonLayer, the map will pan to the geojson for that layer when data is changed, unless the new value is `[]`.


app/extensions/FlyToGeoJsonExtension.ts
```diff
@@ -0,0 +1,47 @@
+import {
+  Layer,
+  LayerExtension,
+  UpdateParameters,
+  FlyToInterpolator,
+  WebMercatorViewport,
+} from '@deck.gl/core';
+import { bbox } from '@turf/bbox';
+import { MultiPolygon, MultiPoint } from '~/gen';
```
TangoYankee, line 9: I think it would be better to use Geometries from [`@types/geojson`](https://www.npmjs.com/package/@types/geojson) rather than the `~/gen`

1) The definitions in the `gen` are cobbled together custom implementations. Meanwhile, `@types/geojson` are well-established
2) We should keep the `FlyToExtension` agnostic to the application it is used within. `@types/geojson` supports this independence (while the gen folder is tied to applications that use the zoning api)

app/extensions/FlyToGeoJsonExtension.ts
```diff
@@ -0,0 +1,47 @@
...
+      throw new Error('Deck instance undefined in FlyToGeoJsonExtension');
+    }
+    if (
+      !Array.isArray(props.data) &&
+      ((Array.isArray(oldProps.data) && oldProps.data.length === 0) ||
+        changeFlags.dataChanged)
+    ) {
+      const viewport = this.context.viewport as WebMercatorViewport;
+      const [minX, minY, maxX, maxY] = bbox(
+        props.data as MultiPolygon | MultiPoint,
```
TangoYankee, line 27: If we swap to `@types/geojson`, we can use the [Geometry type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/865051c6befe8d67f7ac2fdcac6cdf4922c1dc83/types/geojson/index.d.ts#L69), which supports all the shapes- not just the two defined in our OpenAPI spec.

app/components/layers/useCapitalProjectBudgetedGeoJsonLayer.client.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return new GeoJsonLayer<CapitalProjectBudgetedGeoJson["properties"]>({
+    id: "capitalProjectBudgetedGeoJson",
+    data:
+      managingCode === undefined || capitalProjectId === undefined
+        ? []
+        : `${import.meta.env.VITE_ZONING_API_URL}/api/capital-projects/${managingCode}/${capitalProjectId}/geojson`,
+    pickable: true,
+    getFillColor: ({ properties }) => {
+      const { managingCode: selectedProjectManagingCode, id } = properties;
+      switch (`${selectedProjectManagingCode}${id}`) {
```
TangoYankee, line 18: We were more strict than the geojson spec and made the Id a required/guaranteed value in [our geojson endpoints](https://staging-zoning-api.nycplanningdigital.com/#tag/GeoJSON/operation/findCapitalProjectGeoJsonByManagingCodeCapitalProjectId). It's for just this scenario when we want the id already concatenated.

```suggestion
    
    getFillColor: ({ id }) => {
      switch (id) {
```

TangoYankee (changes requested):
Please run `npm run lint`.

1) This will make the quotes consistent
2) It will also reveal that `GeoJsonLayerProps` is defined but not used in the capital project geojson layer

package.json
```diff
@@ -32,6 +32,8 @@
     "@remix-run/node": "^2.7.2",
     "@remix-run/react": "^2.7.2",
     "@remix-run/serve": "^2.7.2",
+    "@turf/bbox": "^7.1.0",
+    "@types/geojson": "^7946.0.14",
```
TangoYankee, line 36: Okay, I know I set you up for this by saying to use `@types/geojson` ....but I think it should be a dev dependency

TangoYankee (approved):

TangoYankee (dismissed):

TangoYankee (approved):

PR #64, opened 2024-08-08
Capital Project Details and Commitments

TangoYankee:
## Description

- Update designs for capital project details
- Add commitments timeline
- Add list of commitment details

closes #17
closes #51

## Bonus
generate kubb files in their own commit

Issue #51, opened 2024-07-15
Update project detail panel to new designs (no commitment data timeline)


TylerMatteo:
## Acceptance Criteria

- [x] Update single project detail panel from #16 to new designs ([desktop](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1187-33573&t=xYpxCPAcg1TPmX2R-0) and [mobile](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1329-23494&t=xYpxCPAcg1TPmX2R-0))

## Out of scope

- Showing the commitments timeline seen in the designs, that will be added as part of implementing commitment-level data.

PR #71, opened 2024-08-14
51/draft/capital project detail

TangoYankee:
# Background
@pratishta I think there is some overlap between #14 and #51. This is a ticket specifically to update the Capital Projects Detail panel to match the latest designs. However, I added some features that might come up while working on #14. I integrated as many changes from that working branch as practical to make future branch integration easier. Also, I added you as a co-author because code from your working branch is in here. I am cleaning up the code and adding some tests. But, I wanted to show the draft and start getting feedback. Some of the features are:

## Features
- The "MobilePanelSizeControl" is split into its own component, meaning it can be reused from the list of projects panel
- The overlay has its padding changed, to allow for a "full width effect" on mobile
- Added support for "exit" and "back" variations of the navigation button
- Added support for if the user clicks a project directly on the map, while they have filtered projects by a district
  - In an in-person discussion, I think we said "navigation should just pop the history stack". However, on further consideration, navigation should conditionally point to a route. For example, we should go back specifically to the list of projects in a district. Otherwise, users who are clicking projects while filtering by district might keep going back to another project when we want really want them to go back to the list of projects.

### Update Aug 16 2024

The code is updated to use the "collapse" chakra transition. It is wrapped around a filler component. This simulates how the commitments are planned to fit into the capital project panel. 

The PR will remain a draft until  PR #64 is on main and those changes can be integrated smoothly into these changes.


app/components/CapitalProjectDetailPanel/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export interface CapitalProjectDetailPanelProps
+  extends MobilePanelSizeControlProps,
+    CapitalProjectDetailProps {}
+
+export function CapitalProjectDetailPanel(
+  props: CapitalProjectDetailPanelProps,
+) {
+  return (
+    <Flex
```
TangoYankee, line 20: `expansion` state definition should probably be lowered into here

app/routes/city-council-districts.$cityCouncilDistrictId.capital-projects_.$managingCode.$capitalProjectId.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [searchParams] = useSearchParams();
+  const { cityCouncilDistrictId } = useParams();
+  const { capitalProject, agencies } = useLoaderData<typeof loader>();
+  const [panelIsFullSize, setPanelIsFullSize] = useState(false);
+
+  const togglePanelIsFullSize = () =>
+    setPanelIsFullSize((panelIsFullSize) => !panelIsFullSize);
+
+  return (
+    <CapitalProjectDetailPanel
```
pratishta, line 46: If we're changing the component name from `CapitalProjectDetailPanel` to `CapitalProjectDetail` we should stay consistent with it

app/components/CapitalProjectDetailPanel/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      borderTopRadius={"base"}
+      borderBottomRadius={{ base: "0", lg: "base" }}
+      background={"white"}
+      direction={"column"}
+      width={{ base: "full", lg: "21.25rem" }}
+      maxW={{ lg: "unset" }}
+      height={{ base: props.isExpanded ? "90vh" : "auto", lg: "auto" }}
```
pratishta: I wonder if we need to specify the non expanded height. From my local machine, a single project view lower on the screen than what's in the designs (which is closer to about half the viewport).

app/components/MobilePanelSizeControl.tsx
```diff
@@ -0,0 +1,31 @@
+import { Box, BoxProps, Hide } from "@nycplanning/streetscape";
+
+export interface MobilePanelSizeControlProps extends BoxProps {
+  isExpanded: boolean;
+  isExpandedToggle: () => void;
+}
+export function MobilePanelSizeControl({
```
pratishta, line 7: Nit: we might want a more descriptive name, just because I couldn't figure out what this actually was meant to be until I looked at the Box component closely. Maybe something like `ExpandBar` or `ResizeBar`.

pratishta (commented):
The updated designs look good. Two things:
* A lot of this work might be easier to integrate with #14 merged, at the very least in terms of testing and QA. Nonetheless, I'm glad to look at the how we wanna have more cohesion in the code, especially before the filter by CD work starts 
* I need a bit more clarity on this

> However, on further consideration, navigation should conditionally point to a route. For example, we should go back specifically to the list of projects in a district. Otherwise, users who are clicking projects while filtering by district might keep going back to another project when we want really want them to go back to the list of projects.

Are you thinking of users who, after they choose a district to filter by, are clicking on those filtered projects on the map? Or those filtered projects from the list? My inclination is to leave it to the user to know that if they're clicking through multiple projects that they will go back to those projects before coming back to the list.

TangoYankee:
> Are you thinking of users who, after they choose a district to filter by, are clicking on those filtered projects on the map?

Yes, my interpretation of the designs is that the navigation icon is an indicator of which panel the user will be sent to. When it's a "close" icon, the user returns the "root" route and all of the content panels are closed. When its a "back arrow" icon, the user will return to the previous panel. This would be the content panel with the list of projects within the active district. 

If the user wants to consecutively select multiple projects directly from the map and then scrub through the literal browser history of projects they looked at, then they can use the native browser history buttons.

app/components/CapitalProjectDetailPanel/index.tsx
TangoYankee: Yeah, these aren't great values. But, things will shift when we add the commitments timeline. I think it makes sense to wait until we add the commitments timeline to reconsider the values.

app/routes/city-council-districts.$cityCouncilDistrictId.capital-projects_.$managingCode.$capitalProjectId.tsx
TangoYankee: I think the way I have been defining and exporting "main" components from their folders might be causing some confusion here.

The top level component is still `CapitalProjectDetailPanel`. However, it is exported from the `index` file of the `CapitalProjectDetailPanel` folder. 

`CapitalProjectDetail` is a sub-component of the `CapitalProjectDetailPanel`. It is  focuses specifically on the `Details` of the project and used only within the `CapitalProjectDetailPanel`.

There is whole discussion around the merits of defining components within the `index` folder. I don't really know how we want to have that discussion.

TangoYankee:
Closing in favor of #76

PR #74, opened 2024-08-15
Update README

### LINKED ISSUE #73: Update ReadMe
- [x] update `README.md` to better explain set up and running the cp-map locally.
- [ ] Add `.vscode` to gitignore so that devs can easily use `git add *` when staging files for commits

horatiorosa:
#### Description
Update `README` to better explain how to set up and run `ae-cp-map` locally.

#### Tickets
Closes #73

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,49 @@
 
 Front end for Application Engineering's capital projects map.
 
-## Dev setup
-- `nvm use`
-- `npm i`
-- `npm run dev`
+# Running the project locally
+
+### Dev setup:
+
+### Ensure you're using the correct version of Node
+The `.nvmrc` file tells you which version of node you should be using to run the project. If you're using [nvm](https://github.com/nvm-sh/nvm) and already have the correct version installed, you can switch by running  
+ ```
+ nvm use
+```
+from the root of this repo.
+
+### Install dependencies
+Once you have cloned this repo, install the necessary dependencies:
+```
+npm i
+```
+
+### Set up your `.env` file to point to a running instance of the [Zoning API](https://github.com/NYCPlanning/ae-zoning-api)
+This project relies on data from a companion API known as the "Zoning API". You will need to set up a `.env` file pointing to a URL where an instance of the Zoning API is hosted. The easiest way to do this is to clone that repo and run it locally. To do that, follow these steps:
+
+1. Go to the [GitHub repo](https://github.com/NYCPlanning/ae-zoning-api) for Zoning API
+2. Follow the steps there to clone the project, run the API, and run its database using docker compose
+3. Leave that Zoning API and its database running in a separate terminal, and navigate back to where you cloned this repo.
+4. Back in your cloned copy of this repo, create a file called `.env` in the root folder of the project and copy the contents of `sample.env` into that new file. The default value in `sample.env` should set the Zoning API URL variable to the correct value for your locally running instance of Zoning API
+
+### Run the project
+Finally, to run this project locally:
+```
+npm run dev
+```
+
+## Pulling data from [Zoning API](https://github.com/NYCPlanning/ae-zoning-api)
+This app relies heavily on data from the Zoning API backend application. It makes use of some front end libraries to make pulling data and managing the state of that data easier:
+
+* [Axios](https://axios-http.com/docs/intro) - A framework-agnostic HTTP client for javascript and typescript applications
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2,10 +2,62 @@
...
+
+> You may see an error in your terminal after running `npm run generate` referring to the "path" argument being undefined. The root cause of this is still being looked into but it doesn't seem to affect the generated code so you can safely ignore it for now.
+
+Unless you think the contents of the Zoning API OpenAPI spec file have changed since the last time the client code was generated, you can safely skip this step.
+
+### Using the Zoning API client code
+Once you have this application configured to point to a running instance of Zoning API, you can start to pull in data. To query data from the Zoning API in our React code, we only need to import and call the hooks generated by Kubb. The `Providers` component makes sure our entire application is wrapped by TanStack Query's `QueryClientProvider` component, so we should be able to call the hooks within any component. To query data, simply import the hook corresponding to the data you need from the `/src/gen` folder and call it within your React component.
+
+```
+import { useGetLandUses } from "./gen" // Update relative import path based on your component's file location
```
pratishta, line 54: I think we have `useFindLandUses` instead of `useGetLandUses` for our generated hooks, but even this is used in `map-poc`. Let's use an example from this repo, like `findCityCouncilDistricts`

README.md
```diff
@@ -2,10 +2,62 @@
...
+
+> You may see an error in your terminal after running `npm run generate` referring to the "path" argument being undefined. The root cause of this is still being looked into but it doesn't seem to affect the generated code so you can safely ignore it for now.
+
+Unless you think the contents of the Zoning API OpenAPI spec file have changed since the last time the client code was generated, you can safely skip this step.
+
+### Using the Zoning API client code
+Once you have this application configured to point to a running instance of Zoning API, you can start to pull in data. To query data from the Zoning API in our React code, we only need to import and call the hooks generated by Kubb. The `Providers` component makes sure our entire application is wrapped by TanStack Query's `QueryClientProvider` component, so we should be able to call the hooks within any component. To query data, simply import the hook corresponding to the data you need from the `/src/gen` folder and call it within your React component.
```
pratishta, line 51: I think the part about us using TanStack Query is no longer true for this repo. I think it's better to just remove this part all together or get others' input to document how we do data fetching here.

pratishta (changes requested):
I also added an extra item to the issue to add `.vscode` to gitignore. I think this PR would be a good place to include that dev flow adjustment

README.md
horatiorosa: Sounds good. We can get @TangoYankee to take a look at his leisure.

TangoYankee: > I think the part about us using TanStack Query is no longer true for this repo. I think it's better to just remove this part all together or get others' input to document how we do data fetching here.


Correct. We use axios to make requests via the loader functions because the loaders execute on the server side. We do still use kubb generated functions to configure the axios calls. An example is: [`findCommunityDistrictsByBoroughId`](https://github.com/NYCPlanning/ae-cp-map/blob/e877149a449da8e51ecbbf45dac67f015210978d/app/root.tsx#L78)

README.md
TangoYankee: Yes, using `find` over `get` is intentional. See this [discussion on names](https://github.com/NYCPlanning/ae-zoning-api/discussions/115#discussion-5982504) (which should find its way into a wiki).

README.md
```diff
@@ -2,10 +2,62 @@
...
+* [Kubb](https://www.kubb.dev/introduction) - A code generation library that reads [OpenAPI specification](https://www.openapis.org/) documents and generates typescript "client" libraries (sometimes also referred to as "SDKs"). It features a set of plugins to generate code for various JS/TS libraries, including Axios and TanStack Query. By pointing Kubb at the [OpenAPI document for Zoning API](https://github.com/NYCPlanning/ae-zoning-api/blob/main/openapi/openapi.yaml), we can avoid writing a lot of boilerplate code.
+
+### Regenerating the Zoning API client code
+When the OpenAPI spec file for Zoning API changes, we will need to regenerate the client code in this repo. To do that, simply run `npm run generate`. Kubb is configured to put all generated code in `/src/gen`. Note that even though that code is generated, we do want to commit it and includes updates in PRs as necessary. A developer should **never** make manually changes or additions to code in the `/src/gen` folder. It should be purely what is outputted by Kubb.
```
TangoYankee, line 44: ```npm run generate``` should be [```npm run kubb:generate```](https://github.com/NYCPlanning/ae-cp-map/blob/e877149a449da8e51ecbbf45dac67f015210978d/package.json#L17)

README.md
```diff
@@ -2,10 +2,62 @@
...
+
+> You may see an error in your terminal after running `npm run generate` referring to the "path" argument being undefined. The root cause of this is still being looked into but it doesn't seem to affect the generated code so you can safely ignore it for now.
```
TangoYankee, line 46: Kubb has since had a few updates. We should no longer see this message.

```suggestion
```

horatiorosa: I didn't notice the change, good to know!

TangoYankee (approved):

horatiorosa:
@pratishta 
I made the change you requested re: the `.gitignore`

pratishta (approved):
🚀

PR #76, opened 2024-08-19
17+51/capital project panel

### LINKED ISSUE #17: Add commitment-level data to project detail panel
Add timeline and commitment-level data to project detail panel.

- [x] Add commitment data to project detail panel according to designs ([desktop](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1194-22664&m=dev) and [mobile](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=1337-20505&m=dev))
- [x] Includes adding timeline and commitments table

Blocked by #51

### LINKED ISSUE #51: Update project detail panel to new designs (no commitment data timeline)

TangoYankee:
# Description
This is a large PR spanning multiple tickets and commits

## Commit 1: (closes #51 )
- Update the design of  Capital Project Panel Details
- Added support for "exit" and "back" variations of the navigation button
- Added support for if the user clicks a project directly on the map, while they have filtered projects by a district
   - In an in-person discussion, I think we said "navigation should just pop the history stack". However, on further consideration, navigation should conditionally point to a route. For example, we should go back specifically to the list of projects in a district. Otherwise, users who are clicking projects while filtering by district might keep going back to another project when we want really want them to go back to the list of projects. 
   - This prompted [a discussion on PR #71 ](https://github.com/NYCPlanning/ae-cp-map/pull/71#issuecomment-2289825708) on whether this is the desired behavior. Input from @TylerMatteo would be helpful.  

## Commit 2:
- generate kubb files 

## Commit 3:
- Fix Admin dropdown tests to use `act`

## Commit 4: (closes #17 )
- Add a commitments table
- Add a commitements timeline
- Use "Collapse" transition from streetscape
   - Upgrade to streetscape 0.11.0
   - Remove properties from 'FilterMenu' that typescript flagged as invalid  
- Integrate styling from City Council District List of Projects
- Refactor Project list to use `Collapse` instead of `Slide`
   - Slide has a animation on mount. This was jarring when moving between an individual project and the list of projects 

closes #17
closes #51


TylerMatteo:
> In an in-person discussion, I think we said "navigation should just pop the history stack". However, on further consideration, navigation should conditionally point to a route. For example, we should go back specifically to the list of projects in a district. Otherwise, users who are clicking projects while filtering by district might keep going back to another project when we want really want them to go back to the list of projects.

Yes this sounds like the correct behavior to me. The left arrow button should always take you back to the list for the district you're filtered to.

app/components/CapitalProjectPanel/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+            />
+          </Flex>
+        </Collapse>
+      </Show>
+      <Show above="lg">
+        <Flex
+          direction={"column"}
+          backgroundColor="gray.50"
+          paddingX={4}
+          paddingBottom={4}
```
TylerMatteo, line 80: As a stop-gap for the vertical centering issue we discussed, you can change this to `paddingBottom={commitmentsAreVisible ? 4 : 0}` and it seems to work as expected (or at least closer than the current version).

TylerMatteo (changes requested):

app/components/CapitalProjectPanel/index.tsx
TangoYankee: updated

TangoYankee:
## Update
Added a max-height and overflow to the desktop version of the capital project detail panel. This will keep it from going off the screen when the panel is longer than a user's screen

TylerMatteo (approved):

Issue #38, opened 2024-07-05
Set Up Matomo Analytics


dhochbaum-dcp:
Matomo Tag Manager Code:
```
<!-- Matomo Tag Manager -->
<script>
  var _mtm = window._mtm = window._mtm || [];
  _mtm.push({'mtm.startTime': (new Date().getTime()), 'event': 'mtm.Start'});
  (function() {
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.async=true; g.src='https://cdn.matomo.cloud/nycplanning.matomo.cloud/container_SmoWWpiD.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
<!-- End Matomo Tag Manager -->
```

PR #97, opened 2024-08-28
38 Added matomo

dhochbaum-dcp:
Adds Matomo tracking, along with an example of a tracked event.

Completes #38

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -44,6 +44,8 @@ import {
 } from "./components/GoToDistrictBtn";
 import { GoToCommunityDistrictBtn } from "./components/GoToDistrictBtn/GoToCommunityDistrictBtn";
 import { WelcomePanel } from "./components/WelcomePanel";
+import { useEffect } from "react";
+import { initializeMatomoTagManager } from "./utils/analytics";
 
 export type BoroughId = null | string;
 export type DistrictType = null | "cd" | "ccd";
@@ -142,6 +144,9 @@ function Document({
 }
 
 export default function App() {
+  useEffect(() => {
+    initializeMatomoTagManager("SmoWWpiD");
+  }, []);
   const navigate = useNavigate();
   const { pathname } = useLocation();
   const [searchParams, setSearchParams] = useSearchParams();
--- a/app/utils/analytics.js
+++ b/app/utils/analytics.js
@@ -1,0 +1,44 @@
+export function analytics(eventData) {
+  if (eventData.value) {
+    window._paq.push([
+      "trackEvent",
+      eventData.category,
+      eventData.action,
+      eventData.name,
+      eventData.value,
+    ]);
+  } else {
+    window._paq.push([
+      "trackEvent",
+      eventData.category,
+      eventData.action,
+      eventData.name,
+    ]);
+  }
+}
+
+export function initializeMatomoTagManager(containerID) {
+  // From the initial script they provide for setup
+  var _mtm = (window._mtm = window._mtm || []);
+  _mtm.push({ "mtm.startTime": new Date().getTime(), event: "mtm.Start" });
+  var d = document,
+    g = d.createElement("script"),
+    s = d.getElementsByTagName("script")[0];
+  g.type = "text/javascript";
+  g.id = "matomo-tag-manager";
+  g.async = true;
+  g.src = `https://cdn.matomo.cloud/nycplanning.matomo.cloud/container_${containerID}.js`;
+  s.parentNode.insertBefore(g, s);
+
+  // To track pageviews, watching for changes to location.pathname
+  let previousUrlPath = location.pathname;
+  const observer = new MutationObserver(function () {
+    if (location.pathname !== previousUrlPath) {
+      previousUrlPath = location.pathname;
+      window._paq.push(["setCustomUrl", location.href]);
+      window._paq.push(["trackPageView"]);
+    }
+  });
+  const config = { subtree: true, childList: true };
+  observer.observe(document, config);
+}
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ node_modules
 /blob-report/
 /playwright/.cache/
  .vscode 
+.DS_Store
--- a/app/components/AdminDropdown/DistrictTypeDropdown.tsx
+++ b/app/components/AdminDropdown/DistrictTypeDropdown.tsx
@@ -1,4 +1,5 @@
 import { AdminDropdownProps, AdminDropdown } from ".";
+import { analytics } from "../../utils/analytics";
 
 export interface DistrictTypeDropdownProps
   extends Pick<AdminDropdownProps, "selectValue"> {
@@ -12,6 +13,11 @@ export function DistrictTypeDropdown({
   const updateDistrictType = (nextDistrictType: string | null) => {
     const nextSearchParams: Record<string, string> =
       nextDistrictType === null ? {} : { districtType: nextDistrictType };
+    analytics({
+      category: "Dropdown Menu",
+      action: "Change District Type",
+      name: nextDistrictType,
+    });
     updateSearchParams(nextSearchParams);
   };
 
--- a/setup-test.ts
+++ b/setup-test.ts
@@ -25,6 +25,11 @@ Object.defineProperty(window, "scrollTo", {
   value: vi.fn(),
 });
 
+Object.defineProperty(window, "_paq", {
+  writable: true,
+  value: [],
+});
+
 /**
  * Mock Service Worker can be configured once the Handlers are generated
  */


app/utils/analytics.js
```diff
@@ -0,0 +1,17 @@
...
+  }
+}
+
+export function initializeMatomoTagManager(containerID) {
+  var _mtm = window._mtm = window._mtm || [];
+  _mtm.push({'mtm.startTime': (new Date().getTime()), 'event': 'mtm.Start'});
+  var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+  g.type='text/javascript';
+  g.id='matomo-tag-manager';
+  g.async=true; g.src=`https://cdn.matomo.cloud/nycplanning.matomo.cloud/container_${containerID}.js`; s.parentNode.insertBefore(g,s);
```
TangoYankee, line 16: It looks like this section is creating a script tag:

```javascript
<script type='text/javascript' id='matomo-tag-manager' async=true src=`https://cdn.matomo.cloud/nycplanning.matomo.cloud/container_${containerID}.js`/>
```

Would it be simpler to move the fully constructed tag to the end of the [document body tags](https://github.com/NYCPlanning/ae-cp-map/blob/6a436307839908147d88cef01674ec59fdf3b086/app/root.tsx#L137)?

app/root.tsx
```diff
@@ -142,6 +144,13 @@ function Document({
 }
 
 export default function App() {
+  let ranEffect = false;
+  useEffect(() => {
+    if (!ranEffect) {
+      initializeMatomoTagManager("SmoWWpiD");
+      ranEffect = true;
+    }
+  }, []);
```
TangoYankee, line 153: The below code should have the same behavior. 

An empty dependency array makes the `useEffect` [run only once](https://react.dev/reference/react/useEffect#examples-dependencies) (on component mount). This should be why `initializeMatomoTagManager` is only running once. 

The `ranEffect` actually is probably trying to rerun the `initializeMatomoTagManager`. This is because it is not wrapped in 'useState'. Instead it is directly defined within the component. Consequently, every time the component re-renders, `ranEffect` is reinitialized and set to `false`. 

Fortunately, the `useEffect` only runs once, regardless of component rerenders, because the dependency array is empty. Consequently, the `if(!ranEffect)` is never run a second time.

If you do see the `initializeMatomoTagManager` run twice in development, it's a `dev mode` feature meant to test for memory leaks. It will truly only run once in production.

```suggestion
  useEffect(() => {
      initializeMatomoTagManager("SmoWWpiD");
  }, []);
```

TangoYankee:
@dhochbaum-dcp , Tyler asked me to review this more closely. But, I need help getting more context. What is the expected outcome of these changes? What are the steps (and permissions) required to validate the changes are having the desired affect? Ie) "we should see X data when user visits Y page"

dhochbaum-dcp:
> @dhochbaum-dcp , Tyler asked me to review this more closely. But, I need help getting more context. What is the expected outcome of these changes? What are the steps (and permissions) required to validate the changes are having the desired affect? Ie) "we should see X data when user visits Y page"

I just sent you an invite to Matomo.

You should see data for the corresponding user's session when 
a) changing the district type in the select
or b) clicking an outbound link

image

You should see network requests to nycplanning.matomo.cloud upon loading the page, and upon performing one of the above.

It does not currently track the URLs properly as you navigate the site - I will investigate and give you an update when that is hopefully fixed, so that you can continue with the review.

app/utils/analytics.js
dhochbaum-dcp: I don't know if there is functionally any difference between the two, but I would prefer to leave the code as-is to maintain consistency with the code snippet they provide, and across our different repos.

TangoYankee:
Thank you for access to the dashboard and the overview on what to look for.

For the page change issue:

> It does not currently track the URLs properly as you navigate the site - I will investigate and give you an update when that is hopefully fixed, so that you can continue with the review. 

As I read [the code](https://github.com/NYCPlanning/ae-cp-map/pull/97/files#diff-8a5ab491182de51ef13fd66206b221bdc90bc420014e23bfd50083a98058de5dR22), we're using the Browser's [`window.location` property](https://developer.mozilla.org/en-US/docs/Web/API/Window/location). I have two hypotheses as to why this might cause issues. 
1) Remix is running the function on the server and there consequently is no `window` object available 
2) Remix is wrapping the `window.location` property in such a way that path changes are not registering (This is a hand-wavy explanation because the exact behavior is abstracted away)

**Tangent**: *The use of `pathname` might also cause us to miss some changes. The `dropdown` changes the query parameters but does not change the url path. I think `pathname` excludes the query parameters. If we want to track path and query parameters, we might need to look at other `location` properties* 

If using the Browser `window.location` property is causing issues, it might be worth trying the [`useLocation`](https://remix.run/docs/en/main/hooks/use-location) remix hook. The `app/utils/analytics.js` file likely won't have access to this hook and so the page tracking would be moved into an actual component. We might be able to do this in the `root.tsx`.

Assuming pathname does give use the path information we want, then we can use the definition we already have at the [`top of root.tsx`](https://github.com/NYCPlanning/ae-cp-map/blob/6dc16acd72acef9f8fed9d70e8ff9524e242bfdd/app/root.tsx#L151) file. We can create an effect that watches for changes and sends subsequent updates.

on line 156, consider adding:
```typescript
useEffect(() => {
    window._paq.push(['trackPageView']);
}, [pathname])
```

This effect should run everytime the `pathname` changes. Let me know if this helps.

dhochbaum-dcp:
> Thank you for access to the dashboard and the overview on what to look for.
> 
> For the page change issue:
> 
> > It does not currently track the URLs properly as you navigate the site - I will investigate and give you an update when that is hopefully fixed, so that you can continue with the review.
> 
> As I read [the code](https://github.com/NYCPlanning/ae-cp-map/pull/97/files#diff-8a5ab491182de51ef13fd66206b221bdc90bc420014e23bfd50083a98058de5dR22), we're using the Browser's [`window.location` property](https://developer.mozilla.org/en-US/docs/Web/API/Window/location). I have two hypotheses as to why this might cause issues.
> 
> 1. Remix is running the function on the server and there consequently is no `window` object available
> 2. Remix is wrapping the `window.location` property in such a way that path changes are not registering (This is a hand-wavy explanation because the exact behavior is abstracted away)
> 
> **Tangent**: _The use of `pathname` might also cause us to miss some changes. The `dropdown` changes the query parameters but does not change the url path. I think `pathname` excludes the query parameters. If we want to track path and query parameters, we might need to look at other `location` properties_
> 
> If using the Browser `window.location` property is causing issues, it might be worth trying the [`useLocation`](https://remix.run/docs/en/main/hooks/use-location) remix hook. The `app/utils/analytics.js` file likely won't have access to this hook and so the page tracking would be moved into an actual component. We might be able to do this in the `root.tsx`.
> 
> Assuming pathname does give use the path information we want, then we can use the definition we already have at the [`top of root.tsx`](https://github.com/NYCPlanning/ae-cp-map/blob/6dc16acd72acef9f8fed9d70e8ff9524e242bfdd/app/root.tsx#L151) file. We can create an effect that watches for changes and sends subsequent updates.
> 
> on line 156, consider adding:
> 
> ```ts
> useEffect(() => {
>     window._paq.push(['trackPageView']);
> }, [pathname])
> ```
> 
> This effect should run everytime the `pathname` changes. Let me know if this helps.

Thanks for looking into this - the event was actually properly firing when the page changed, but it continued to record the url of the initially loaded page.  I resolved this by manually setting the URL prior to firing the pageview.  It's now working how we would like - changing the URL parameters does not trigger a pageview, and we will manually track events; and changing the path does trigger a pageview.

Looks ready to move forward with review.

TangoYankee:
> I resolved this by manually setting the URL prior to firing the pageview. It's now working how we would like - changing the URL parameters does not trigger a pageview, and we will manually track events; and changing the path does trigger a pageview.

Ahh, good to know. 

It seems my code suggestions are merely stylistic differences and I'm not going to hold an approval over them. I would like to put the changes onto the QA environment so that I can sanity check the code working in a deployed environment.

TangoYankee (changes requested):
I do see the target metrics populating in the Matomo dashboard:
- Changing the district type
- Change the application path
- Clicking to the external links to bytes and omb

However, linting the code (`npm run lint`) revealed the following errors:
```
/home/wsl2user/planning/ae-cp-map/app/utils/analytics.js
   3:5   error  '_paq' is not defined                  no-undef
  11:5   error  '_paq' is not defined                  no-undef
  35:51  error  'mutations' is defined but never used  no-unused-vars
  38:7   error  '_paq' is not defined                  no-undef
  39:7   error  '_paq' is not defined                  no-undef
```

horatiorosa:
I also ran into those `_paq` errors. I think I resolved them by changing them to `window._paq`, though I am not sure that is the right way to go.

dhochbaum-dcp:
> * I do see the target metrics populating in the Matomo dashboard:
>   
>   * Changing the district type
>   * Change the application path
>   * Clicking to the external links to bytes and omb
> * However, linting the code (`npm run lint`) revealed the following errors:
> 
> ```
> /home/wsl2user/planning/ae-cp-map/app/utils/analytics.js
>    3:5   error  '_paq' is not defined                  no-undef
>   11:5   error  '_paq' is not defined                  no-undef
>   35:51  error  'mutations' is defined but never used  no-unused-vars
>   38:7   error  '_paq' is not defined                  no-undef
>   39:7   error  '_paq' is not defined                  no-undef
> ```

These have been resolved.

TangoYankee (changes requested):
Thank you for resolving the lint errors. 

I double checked the tests. `npm run test` is successful. However, `npm run vitest` results in a failure.

```
TypeError: Cannot read properties of undefined (reading 'push')
 ❯ Module.analytics app/utils/analytics.js:11:17
      9|     ]);
     10|   } else {
     11|     window._paq.push([
       |                 ^
     12|       "trackEvent",
     13|       eventData.category,
```

My hunch is that `_paq` might need to be added to the [existing mock of the window object](https://github.com/NYCPlanning/ae-cp-map/blob/13bc33211f0a32abaedafe01b97110e5d536552d/setup-test.ts#L8)

dhochbaum-dcp:
> Thank you for resolving the lint errors.
> 
> I double checked the tests. `npm run test` is successful. However, `npm run vitest` results in a failure.
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'push')
>  ❯ Module.analytics app/utils/analytics.js:11:17
>       9|     ]);
>      10|   } else {
>      11|     window._paq.push([
>        |                 ^
>      12|       "trackEvent",
>      13|       eventData.category,
> ```
> 
> My hunch is that `_paq` might need to be added to the [existing mock of the window object](https://github.com/NYCPlanning/ae-cp-map/blob/13bc33211f0a32abaedafe01b97110e5d536552d/setup-test.ts#L8)

Thanks for catching that, I've pushed the fix.

TangoYankee:
Okay, the changes look good. I did some cleanup work to 1) rebase off main 2) lint the code 3) consolidate the work into a single commit and 4) add meta data into the commit message. Do you mind if I force push this history change? You would still be listed as the commit author for the work.

TangoYankee (approved):

PR #98, opened 2024-08-29
59 Add Fullstory analytics

horatiorosa:
#### Description
Add Fullstory to CP map 

#### Notes
tracking works (tested in Safari browser as Chrome browsing didn't reflect in FS segment)

~I do have an error I cannot fix as yet; `FullStory namespace conflict. Please set window["_fs_namespace"].`~
The above error shouldn't occur in prod. Only shows up in develop bc of the `useEffect` effect. 

#### Tickets
Closes # 59

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -44,8 +44,11 @@ import {
 import { GoToCommunityDistrictBtn } from "./components/GoToDistrictBtn/GoToCommunityDistrictBtn";
 import { WelcomePanel } from "./components/WelcomePanel";
 import { useEffect } from "react";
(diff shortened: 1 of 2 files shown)


horatiorosa:
There is a test segment set up in fullstory `cp-map test (localhost:5173)`

TangoYankee:
> I do have an error I cannot fix as yet; FullStory namespace conflict. Please set window["_fs_namespace"].
The above error shouldn't occur in prod. Only shows up in develop bc of the useEffect effect.

I'm keying into the "shouldn't". Have we tried locally to create and run a production version of the build?

TangoYankee (approved):

Issue #101, opened 2024-09-06
Bug: Project list updates when new district selected from filter menu


pratishta:
Steps to reproduce:
1. Select City Council District on filter menu
2. Select district number 1 on filter menu
3. Hit 'Go to selected district'
4. Go to second page of project list
5. Select district number 2 on filter menu (do not hit 'Go to selected district)

The project list should update to show projects from the first page of district 1. My thoughts are that the project list should not change unless the user actually clicks 'Go to selected district' button

PR #108, opened 2024-09-13
Project list updates when filter menu changes

pratishta:
Closes https://github.com/NYCPlanning/ae-cp-map/issues/101

The issue stems from the fact that when we change something in the dropdowns, we don't consider the page query parameter. Or as Tim mentioned, any previous query parameter. 

~~There are two potential solutions:~~

~~Option 1: Commit c245ed9ddca5ca382f2989f0ce8a2a3842f92b0e passes the page number to each dropdown component so it's accounted for when a value is changed.~~
~~Option 2: Commit ab0d209db97ed286503f55afa50f28d4ce88f769 explicitly deletes the previous page param and appends the first page upon every 'go to district'.~~

~~Both methods explicitly have the first page as a query param in the url whereas before we didn't have it at all. I'm partial to the second method because it's shorter and more straightforward.~~

### DIFF
--- a/app/components/AdminDropdown/BoroughDropdown.test.tsx
+++ b/app/components/AdminDropdown/BoroughDropdown.test.tsx
@@ -14,7 +14,7 @@ describe("BoroughDropdown", () => {
     const updateSearchParams = vi.fn();
     render(
       <BoroughDropdown
-        updateSearchParams={updateSearchParams}
+        setAdminParams={updateSearchParams}
         boroughs={boroughs}
       />,
     );
@@ -28,7 +28,7 @@ describe("BoroughDropdown", () => {
     const firstBoroughId = boroughs[0].id;
     render(
       <BoroughDropdown
-        updateSearchParams={updateSearchParams}
+        setAdminParams={updateSearchParams}
         selectValue={firstBoroughId}
         boroughs={boroughs}
       />,
@@ -37,6 +37,8 @@ describe("BoroughDropdown", () => {
     await act(() => userEvent.selectOptions(screen.getByRole("combobox"), ""));
     expect(updateSearchParams).toHaveBeenCalledWith({
       districtType: "cd",
+      boroughId: null,
+      districtId: null,
     });
   });
 
@@ -45,7 +47,7 @@ describe("BoroughDropdown", () => {
     const firstBoroughId = boroughs[0].id;
     render(
       <BoroughDropdown
-        updateSearchParams={updateSearchParams}
+        setAdminParams={updateSearchParams}
         boroughs={boroughs}
       />,
     );
@@ -56,6 +58,7 @@ describe("BoroughDropdown", () => {
     expect(updateSearchParams).toHaveBeenCalledWith({
       districtType: "cd",
       boroughId: firstBoroughId,
+      districtId: null,
     });
   });
 });
--- a/app/utils/utils.ts
+++ b/app/utils/utils.ts
@@ -1,4 +1,5 @@
 import { compareAsc, format, getMonth, getYear } from "date-fns";
+import { SearchParamChanges } from "./types";
 
 const getFiscalYearForDate = (date: Date): number => {
   const year = getYear(date);
@@ -17,18 +18,19 @@ export const currentDate = () => {
 };
 
 // from https://www.jacobparis.com/content/remix-pagination
-export function setNewSearchParamsString(
+export function setNewSearchParams(
   searchParams: URLSearchParams,
-  changes: Record<string, string | number | undefined>,
+  changes: SearchParamChanges,
 ) {
   const newSearchParams = new URLSearchParams(searchParams);
+
   for (const [key, value] of Object.entries(changes)) {
-    if (value === undefined) {
+    if (value === (undefined || null)) {
       newSearchParams.delete(key);
       continue;
     }
     newSearchParams.set(key, String(value));
   }
 
-  return newSearchParams.toString();
+  return newSearchParams;
 }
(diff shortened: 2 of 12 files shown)


TangoYankee (changes requested):
I think there is a strategy where updating the geography dropdown only updates its own parameters, without dropping other parameters- regardless of what those other parameters are. For this we can update the 'updateSearchParams' function that is defined in the root and then passed to the dropdown components.

The `updateSearchParams` function is defined in the root and is a wrapper around [useSearchParams](https://remix.run/docs/en/main/hooks/use-search-params). We then pass this function to the dropdown components. Right now, we update the query parameters without regard to the previous url parameters. However, `setSearchParams` is aware of the previous url parameters. We should be able to modify `updateSearchParams` to take the `nextUrlParams` and merge them with the previousUrlParams, regardless of what those other parameters are.

pratishta:
@TangoYankee 

> I think there is a strategy where updating the geography dropdown only updates its own parameters, without dropping other parameters- regardless of what those other parameters are. 

Yes. Updating a filter menu drop down currently does not impact any parameter other than what one that dropdown is associated with. Borough dropdown = `boroughId` param. Are you seeing that this is not the case with either of my proposed options?

TangoYankee:
> currently does not impact any parameter other than what one that dropdown is associated with

The dropdown does a hard overwrite of all the query parameters. Any parameters that it does not reset are simply lost. ...this was a poor design choice on my part when first implementing the dropdowns. I think this choice will continue to cause trouble if we don't think of a general solution.

> we want the solution in the PR to not be specific to just the `page` param?

Yes, I think we can get a solution that preserves any unrelated parameters

pratishta:
Updated PR to use [setSearchParams](https://remix.run/docs/en/main/hooks/use-search-params#setsearchparamsprevparams--newparams-navigateoptions) to combine the incoming search params with previous params. We still need to reset to page 1 when going to a new district though, which is done through using `.delete`.

pratishta:
Took into consideration @TangoYankee's feedback from [his branch](https://github.com/NYCPlanning/ae-cp-map/pull/111/files). Namely, using `AdminParams` and adopting the philosophy of including necessary params instead of deleting unnecessary params. 

I also moved some types to the `util` folder. I also realized that we already had away to update / merge params with `setNewSearchParamsString` so I changed it to be a bit more generic so we can use it beyond Pagination.

app/utils/utils.ts
```diff
@@ -17,18 +18,19 @@ export const currentDate = () => {
 };
 
 // from https://www.jacobparis.com/content/remix-pagination
-export function setNewSearchParamsString(
+export function setNewSearchParams(
   searchParams: URLSearchParams,
-  changes: Record<string, string | number | undefined>,
+  changes: Record<string, string | number | undefined | null> | AdminParams,
```
TangoYankee, line 23: I see a few instances where we [union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html) `Record<string, string>` with `AdminParams`. My perception is we're doing this because `AdminParams` are incompatible with `Record<string, string>`. *(And also because district types are an odd case. details on how to resolve this case at the end of the list of suggestions.)* This is because AdminParams have the shape `Record<string, string | null>`. The `null` makes it incompatible.

The union does resolve this conflict. However, I think it obfuscates the intent of the typing. If we create a new type, I think we can make it compatible with AdminParams and clarify the intended types.

1. New `SearchParamChanges` type:
```typescript
export type SearchParamChanges = Record<string, string | number | null | undefined> 
``` 

2. We can then use this type in the high-level functions:
- `app/utils/utils.ts line 21`
```typescript
export function setNewSearchParams(
   searchParams: URLSearchParams
   changes: SearchParamChanges
)
```
- `app/root.tsx line 172`
```typescript
const updateSearchParams = (
    nextSearchParams: SearchParamChanges,
```

3. And then in the dropdowns, AdminParams should now be compatible with `SearchParamChanges` and we can specify that the dropdown params must meet the `AdminParams` criteria.
- `app/components/AdminDropdown/BoroughDropdown.tsx line 8`
```typescript
setAdminParams: (value: AdminParams) => void;
```
- and all the other dropdowns, too.
- Adding [these value checks](https://github.com/NYCPlanning/ae-cp-map/pull/108/files#diff-4c2ea46ba29c278a05289ff73d08b954319ea5e8950e56c8f226335791a3ff75) to the district type dropdown, to ensure the new value conforms to the AdminParams type.

TangoYankee (changes requested):
The foundation looks good and I like reusing the search param update utility function. I have a couple suggestions related to testing and typing.

app/utils/utils.ts
TangoYankee: Mmmm, I'm trying this myself and it's giving me a hard time 🙃

pratishta: Yeah so admittedly, I popped the union in there to just get the types to work lol But-- what I had in mind was to make it so that we didn't write functions around only admin params while still having the compact and readable benefit `AdminParams` provide in the dropdowns. I agree with having _something_ like SearchParamChanges like maybe even keeping `Record<string, string | number | null | undefined>` as is instead of a new type?

app/components/AdminDropdown/BoroughDropdown.test.tsx
```diff
@@ -14,7 +14,7 @@ describe("BoroughDropdown", () => {
     const updateSearchParams = vi.fn();
```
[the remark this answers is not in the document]
pratishta: Oops forgot about tests, ty

app/utils/utils.ts
TangoYankee: Okay, it's seems to be from mixing `interfaces` and `types`. Changing `AdminParams` from an `interface` to a type seems to resolve the typing problems. I'm updating the above comment to include this change.

pratishta: Great, updated with `SearchParamChanges`

TangoYankee (approved):
🚀  🌔

PR #129, opened 2024-12-20
125 Zooming at max zoom pans map

### LINKED ISSUE #125: Zooming at max zoom pans map

dhochbaum-dcp:
Fixed issue where after fully zooming out using the mouse scroller, any more scrolls move the map to the left/down.

The default minZoom for the MapViewState is 0, so the map was attempting to zoom out, only to be manually corrected in the onViewStateChange.  By setting the minZoom to be MIN_ZOOM in the initial state, we override the default and make the zoom stop before the manual correction in onViewStateChange would otherwise kick in.

Closes #125

### DIFF
--- a/app/components/atlas.client.tsx
+++ b/app/components/atlas.client.tsx
@@ -15,8 +15,8 @@ import {
 } from "./layers";
 import type { MapView, MapViewState } from "@deck.gl/core";
 
-const MAX_ZOOM = 20;
-const MIN_ZOOM = 10;
+export const MAX_ZOOM = 20;
+export const MIN_ZOOM = 10;
 
 const INITIAL_VIEW_STATE = {
   longitude: -74.0008,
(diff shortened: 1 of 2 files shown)


TangoYankee (changes requested):
The issue goes away for the home page. But, I get the behavior on pages that included a "zoom to", ie) `/boroughs/1/community-districts/01/capital-projects?districtType=cd&boroughId=1&districtId=01`

dhochbaum-dcp:
> The issue goes away for the home page. But, I get the behavior on pages that included a "zoom to", ie) `/boroughs/1/community-districts/01/capital-projects?districtType=cd&boroughId=1&districtId=01`

Thanks, good catch.  I've fixed that as well.

TylerMatteo (approved):

TylerMatteo (approved):

TylerMatteo (approved):

Issue #126, opened 2024-12-17
User Story: Search by Attribute - Managing Agency


jessicashanshanhuang:
# User Story: Search by Attribute - Managing Agency

As a member of the general public, I would like to be able to **search for a project by managing agency to compare related capital projects**. 

![Image](https://github.com/user-attachments/assets/a200d283-d3df-431b-ae2e-96ad87c2c267)

## Acceptance Criteria

- A user should be able to go to https://capitalprojects.nycplanningdigital.com/ and navigate to the filter dropdown
- Users should be able to select from a range of managing agencies and choose one agency for as their primary choice 
- Project List component should populate projects related to selected managing agency

## Polish Features
- Have map highlight points and polygons associated by selected managing agency
- Project type tags within a Capital Project Card link back to the filtered Project List

jessicashanshanhuang:
Updating the user story to correspond with [User Story: Resetting Filters](https://github.com/NYCPlanning/ae-cp-map/issues/139)

- Users should have the ability to reset selection for managing agency

![Image](https://github.com/user-attachments/assets/82ae7511-5b08-403e-a6c0-7a96f5dbb9d1)

dhochbaum-dcp:
Hi @jessicashanshanhuang, some questions about the [updated designs](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=3032-3512&p=f&t=4FV6MYbmV3dg1EmU-0):

I don't see a preview of the expanded Managing Agency select.  Should it list the just names of agencies, or should it be something like "Agency Name (Acronym)"?  Should it be sorted alphabetically by agency name, acronym, or another way?

The designs show a small circle with an "x" when a Managing Agency or Project Type is selected.  Can you please confirm that clicking those should clear the selection for the associated field; and confirm it should appear only on Managing Agency and Project type (and no other selects)?

jessicashanshanhuang:
Hey @dhochbaum-dcp,  great questions 🤔

> I don't see a preview of the expanded Managing Agency select. Should it list the just names of agencies, or should it be something like "Agency Name (Acronym)" Should it be sorted alphabetically by agency name, acronym, or another way?

The list should be sorted alphabetically and presented agency name like "Agency Name (Acronym)" `Department of City Planning (DCP)`


> The designs show a small circle with an "x" when a Managing Agency or Project Type is selected. Can you please confirm that clicking those should clear the selection for the associated field; 

[...]

PR #149, opened 2025-02-03
141 Add Managing Agency Filter

dhochbaum-dcp:
~@TylerMatteo here's what I have so far.  You can see in useCapitalProjectsLayer.client.tsx that it's filtering on `managingAgency`, and when you use the Managing Agency dropdown the map results get filtered, but the filtering does not appear to be correct, and when I left off on Thursday I was stuck trying to figure out what exactly it was filtering.~

## Acceptance Criteria

- [x] Add new "Search by Attribute" section on mobile and desktop ([designs](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=3032-3512&p=f&t=ZwA2AWbkW51khlQi-0))
- [x] Add managing agency dropdown, populated with `/agencies` endpoint. 
- [x] "Search" button is disabled until user selects an agency
- [x] When user clicks "Search", projects on map and in list are filtered to only those belonging to the given agency. Query parameter for agency should be reflected in browser query params

Additional changes had to be made to meet the above and match the new designs:
- There is one Search button for all of the filters
- All searches go to the new "capital-projects" route, regardless of District
- All capital project results go to the "capital-projects/:managingCode/:capitalProjectId" route

Outstanding problems that require creation of a new issue:
- Total Projects (TBD) and "as of" date (#151) do not have data yet
- CSVs need to be redone somehow - [working demo of dynamically-generated csv downloads matching user query](https://github.com/NYCPlanning/ae-zoning-api/compare/dh-dynamic-csv-demo)

Things I just noticed on Friday afternoon right before retro:
 - Need to shorten the text area in the select so that the clear selection button no longer covers the ellipsis
 - CD and CCD outlines don't appear until after search button is clicked
 - Results pane not appearing
 - If you select Managing Agency, hit Search, map filters properly.  If you clear the filter, the map doesn't change, and the Search button is disabled.


TylerMatteo:
@dhochbaum-dcp can you expand upon what you mean by the filtering not appearing to be correct? It seems to work for me once you pick an agency but it could be filtering projects it shouldn't and I wouldn't know. Is there a particular project you're expecting to see rendered but aren't?

app/components/AdminDropdown/AgencyDropdown.tsx
```diff
@@ -0,0 +1,64 @@
...
+    nextManagingAgencyAcronym: ManagingAgencyAcronym,
+  ) => {
+    analytics({
+      category: "Dropdown Menu",
+      action: "Change Managing Agency",
+      name: nextManagingAgencyAcronym,
+    });
+
+    const projectType = searchParams.get("projectType") as ProjectTypeCode;
+    const min = searchParams.get("min") as AmountMin;
```
TylerMatteo, line 33: Related to another comment in the `utils/types` file, these are out of scope, no?

app/root.tsx
```diff
@@ -105,6 +127,8 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
         boroughs,
         communityDistricts,
         cityCouncilDistricts: null,
+        agencies,
+        // projectTypes,
```
TylerMatteo, line 131: Out of scope, don't need to leave in placeholders for future work.

app/components/layers/useCapitalProjectsLayer.client.tsx
```diff
@@ -79,8 +92,10 @@ export function useCapitalProjectsLayer() {
       });
     },
     updateTriggers: {
-      getFillColor: [managingCode, capitalProjectId],
-      getPointColor: [managingCode, capitalProjectId],
+      getFillColor: [managingCode, capitalProjectId, managingAgency],
```
TylerMatteo, line 95: Removing `managingAgency` here and on the following line doesn't seem to break anything and would cut down on how many times these functions are executed.

app/components/AdminDropdown/AgencyDropdown.tsx
TylerMatteo, line 1: Can you add some unit tests for this component? If you're not sure what to test for, I can try to help.

app/components/AdminDropdown/AgencyDropdown.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const max = searchParams.get("max") as AmountMax;
+
+    setAttributeParams({
```
TylerMatteo, line 36: There seems to be an issue with this implementation and/or unclear requirements that wouldn't be obvious given the filtering not working on your machine:

On mine, when I selected a new Agency from the dropdown, the project shown on the map immediately filter to only the projects for that new agency. The designs and Jessica's diagrams suggest that shouldn't happen until the user actually clicks "Search" (or else I'm not sure what the point of that button is). I left a comment [here](https://www.figma.com/board/yNPqYRlDy0ruFMLocAZrCb/CPP-System-Flows?node-id=295-514&t=RRGzHkZQHXufIATn-4) for @jessicashanshanhuang to confirm.

I understand this has implications for how we handle query params in general, including those for district types and districts. Assuming the behavior is supposed to work how I'm describing, we can store the state of each dropdown within each component and only update the query params when the user clicks search. I know this means the current dropdown state isn't reflected 1:1 in the params but I'm not sure that's actually a bad thing in this situation.

app/gen/axios/findAgencies.ts
```diff
@@ -12,6 +12,7 @@ export async function findAgencies(
   const res = await client<FindAgenciesQueryResponse>({
     method: "get",
     url: `/agencies`,
+    baseURL: "https://zoning.planningdigital.com/api",
```
TylerMatteo, line 15: Do you know why this `baseURL` property is getting set as a result of your changes? I would expect it to get set as a result of changes to `kubb.config.ts`, which doesn't show up in the diff.

TylerMatteo (changes requested):

TylerMatteo:
@dhochbaum-dcp there's also an issue with no projects show up until you select an agency but I know you can't really fix that with the issues you're having locally. I'll see if I can resolve it and push a change.

app/components/CapitalProjectsList/CapitalProjectsList.tsx
```diff
@@ -40,14 +40,16 @@ export const CapitalProjectsList = ({
   return (
     <>
       <Box paddingBottom={4}>
-        <Text as={"span"}>Mapped Capital Projects</Text>
+        <Text as={"span"}>
+          Mapped Capital Projects as of <strong>xx/xx/xxxx</strong>
```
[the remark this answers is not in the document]
dhochbaum-dcp: Done.

package.json
```diff
@@ -28,9 +28,9 @@
     "@emotion/styled": "^11.11.0",
     "@kubb/swagger-client": "^2.28.3",
     "@nycplanning/streetscape": "^0.12.0",
-    "@remix-run/node": "^2.13.1",
-    "@remix-run/react": "^2.13.1",
-    "@remix-run/serve": "^2.13.1",
+    "@remix-run/node": "^2.15.2",
```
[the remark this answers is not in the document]
dhochbaum-dcp: Probably not, it was probably done while trying to fix the filter issue.  It's been reverted.

app/utils/types.ts
```diff
@@ -1,13 +1,24 @@
 export type BoroughId = null | string;
 export type DistrictType = null | "cd" | "ccd";
 export type DistrictId = null | string;
+export type ManagingAgencyAcronym = null | string;
+export type ProjectTypeCode = null | string;
+export type AmountMin = null | number;
```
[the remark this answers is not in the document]
dhochbaum-dcp: It is out of scope, they have been removed.

app/components/layers/useCapitalProjectsLayer.client.tsx
```diff
@@ -79,8 +92,10 @@ export function useCapitalProjectsLayer() {
...
+      getPointColor: [managingCode, capitalProjectId, managingAgency],
+      getFilterCategory: [managingAgency],
```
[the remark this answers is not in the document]
dhochbaum-dcp: Deleted.

app/components/AdminDropdown/AgencyDropdown.tsx
dhochbaum-dcp: It is out of scope, they have been removed.

app/root.tsx
dhochbaum-dcp: Removed.

app/components/layers/useCapitalProjectsLayer.client.tsx
dhochbaum-dcp: Done.

app/gen/axios/findAgencies.ts
dhochbaum-dcp: I do not.

app/components/AdminDropdown/AgencyDropdown.tsx
dhochbaum-dcp: Added.

app/components/AdminDropdown/AgencyDropdown.tsx
dhochbaum-dcp: I see her response, and followed with a question of my own.  With the way I had to rework things, this is also true of changing District Number as well, I think we should merge as-is and correct the problem once we've determined we're sure that's how we want it to work.

dhochbaum-dcp:
I just pushed a working fix for the map to show capital projects when no filters are selected.  On Monday, I'll update it to pull the list of agencies automatically.

app/gen/axios/findAgencies.ts
TylerMatteo: I regen'd and confirmed this doesn't seem to be due to any changes in this PR. My best guess is its related to [this commit upgrading Kubb](https://github.com/NYCPlanning/ae-cp-map/commit/dc97115b0b12512d0b4c4a9d606c418378405b35). I regen'd at a commit prior to that one and `baseURL` isn't getting set. I think what's happening is the newer version of Kubb is [pulling this url from the OAS spec file](https://github.com/NYCPlanning/ae-zoning-api/blob/main/openapi/openapi.yaml#L7). However, when we pass `baseURL` in our application code using the env var, that still takes precedence so this should be harmless. Resolving - cc @TangoYankee just so it's on your radar.

TylerMatteo:
@dhochbaum-dcp once you push that update, please deploy this to QA and let @jessicashanshanhuang know so she can do some design QA. Jessica - this PR should cover #126 but also generally refactors the whole left-hand panel to match the [new designs](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=3040-8347&t=LmjD0sUofscGzvrL-0), just without the type filter, amount filter, or "reset selection" button. Hoping we can QA that here before moving on to the other filters but let me know if that doesn't work for you.

dhochbaum-dcp:
@jessicashanshanhuang It's up on [QA](https://qa-capitalplanning.nycplanningdigital.com/).

package.json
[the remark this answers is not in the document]
TylerMatteo: Did you forget to push this? It's still showing up in the diff. I'm cool with including the Deck upgrades here as they are necessary to complete the functionality.

app/components/AdminDropdown/AgencyDropdown.tsx
TylerMatteo: Following up on the [comment I left in Figma](https://www.figma.com/board/yNPqYRlDy0ruFMLocAZrCb/CPP-System-Flows?node-id=302-368&t=vzvwcCTq7rto7aKt-0), please go ahead with having the filter apply when "Search" is clicked. If this change necessitates any downstream changes to existing code, please let me know and we can discuss.

app/components/layers/useCapitalProjectsLayer.client.tsx
```diff
@@ -1,44 +1,57 @@
...
-  } else if (onCapitalProjectsInCommunityDistrictPath) {
-    endpointPrefix = `boroughs/${layoutRoute.params.boroughId}/community-districts/${layoutRoute.params.communityDistrictId}/`;
+  if (districtType === "ccd" && districtId) {
+    endpointPrefix = `city-council-districts/${districtId}/`;
+  } else if (districtType === "cd" && boroughId && districtId) {
+    endpointPrefix = `boroughs/${boroughId}/community-districts/${districtId}/`;
   }
 
-  return new MVTLayer<CapitalProjectProperties>({
+  const loaderData = useLoaderData<
```
TylerMatteo, line 35: You're on the right track here using the entire list when no agency is selected but I think we can simplify this logic significantly and remove the need for `"placeholder"` magic strings. I have a solution for this that appears to be working. I'll leave line-specific changes below.

app/components/layers/useCapitalProjectsLayer.client.tsx
```diff
@@ -79,8 +97,15 @@ export function useCapitalProjectsLayer() {
...
     updateTriggers: {
-      getFillColor: [managingCode, capitalProjectId],
+      getFillColor: [managingCode, capitalProjectId, managingAgency],
       getPointColor: [managingCode, capitalProjectId],
+      filterCategories: [managingAgency],
+      extensions: [managingAgency],
     },
+    extensions: [
+      new DataFilterExtension({
+        categorySize: 2,
```
TylerMatteo, line 107: The above changes mean we can change this back to `categorySize: 1`

app/components/layers/useCapitalProjectsLayer.client.tsx
```diff
@@ -47,6 +60,11 @@ export function useCapitalProjectsLayer() {
     autoHighlight: true,
     highlightColor: [129, 230, 217, 218],
     pickable: true,
+    getFilterCategory: (f: Feature<Geometry, CapitalProjectProperties>) => [
+      f.properties.managingAgency,
+      "placeholder",
+    ],
+    filterCategories: filterCategories,
```
TylerMatteo, line 67: This can be set to either the acronym list when the query param is `null` (not including any added values like `placeholder`) or an array containing on the selected agency when the query param isn't null. The simpler logic also means we don't need as much logic and intermediate variables in lines 35-49. This is what I set it to:
```
filterCategories:
      managingAgency === null ? fullAgencyAcronymList : [managingAgency],
```

app/components/layers/useCapitalProjectsLayer.client.tsx
```diff
@@ -47,6 +60,11 @@ export function useCapitalProjectsLayer() {
     autoHighlight: true,
     highlightColor: [129, 230, 217, 218],
     pickable: true,
+    getFilterCategory: (f: Feature<Geometry, CapitalProjectProperties>) => [
```
TylerMatteo, line 63: We don't want to pollute the value here to include anything not in the tile itself if we can help it. With the rest of the suggestings here, I was able to set this back to:
```
getFilterCategory: (f: Feature<Geometry, CapitalProjectProperties>) =>
      f.properties.managingAgency,
```

app/components/layers/useCapitalProjectsLayer.client.tsx
```diff
@@ -79,8 +97,15 @@ export function useCapitalProjectsLayer() {
       });
     },
     updateTriggers: {
```
TylerMatteo, line 99: Because `DataFilterExtension` filters on on the GPU, not the CPU, it doesn't rely on the update loop that `updateTriggers` manages. I have this set to
```
updateTriggers: {
      getFillColor: [managingCode, capitalProjectId],
      getPointColor: [managingCode, capitalProjectId],
    },
```
and the filtering appears to be working with the rest of the changes outlined here.

TylerMatteo (dismissed):

package.json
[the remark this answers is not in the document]
dhochbaum-dcp: No idea what happened there, pushing an update now.

app/components/layers/useCapitalProjectsLayer.client.tsx
dhochbaum-dcp: @TylerMatteo RE: AgencyDropdown, we'll need to discuss further.  Currently, all changes to the UI are made instantaneously.  I'm unclear why we would want it to work any other way, and I'm not sure how an implementation to wait for a user to click "Search" would work.  

RE: The "placeholder" magic strings - I set it up that way to make sure that we had a solution that would work with multiple filters, because we are about to add Project Type.  If you'd like, I can make the changes you've suggested, and then re-implement my design once we've added Project Type, or we can leave it as is.

TylerMatteo: > RE: AgencyDropdown, we'll need to discuss further. Currently, all changes to the UI are made instantaneously. I'm unclear why we would want it to work any other way, and I'm not sure how an implementation to wait for a user to click "Search" would work.

We want it to apply the filter when the user clicks "search" because that's how Jessica confirmed it is expected to work. As for how this would work, well have to manage the state of the dropdown in a way where the selected value isn't immediately "synced" with the query params. You could try storing the state for the dropdown with `useState` in a component that also has access to the "Search" button. Then, when the user hits Search, you can update the query param accordingly.

> RE: The "placeholder" magic strings - I set it up that way to make sure that we had a solution that would work with multiple filters, because we are about to add Project Type. If you'd like, I can make the changes you've suggested, and then re-implement my design once we've added Project Type, or we can leave it as is.

Please make the changes I suggested and whoever takes on implementing other filters can build on your work (or you can reimplement with real values if it's you)

dhochbaum-dcp: Ok, this update has been made.

dhochbaum-dcp:
@TangoYankee yes, showing `totalProjects` is just a placeholder, because it is not in the API yet.

TylerMatteo:
@TangoYankee I'm working on a feature flag implementation, stand by.

TangoYankee:
How do we want to handle cases where users starts, but do not finish, selecting a district geography?  Right now, it shows results on the map but not in the list.

[unfinished_district_selection.webm](https://github.com/user-attachments/assets/b08096e5-11cb-4ba5-8b14-6ccd9f8a051d)

TangoYankee:
> How do we want to handle cases where users starts, but do not finish, selecting a district geography? Right now, it shows results on the map but not in the list.
> 

From an in-person discussion, we're going to disable search when the district is in an intermediate state of selection

TangoYankee:
We always say "Mapped Capital Projects" in the project list. To clarify, when the filter doesn't include a geography, the results also includes unmapped projects. I don't know that it's a bug but it does feel like it's misleading. Also, when selecting a unmapped project after selecting a mapped project, the zoom stays on the mapped project. 

[handling_unmapped_projects.webm](https://github.com/user-attachments/assets/a5e7bf6b-bc8a-49f0-a42c-56de3b411b33)

TylerMatteo:
That is a good call out about unmapped projects - honestly I think there's an implicit requirement to _always_ filter out unmapped projects. I'll talk to Jessica about it but I think we can go forward with this implementation considering it's behind a feature flag.

app/routes/city-council-districts.$cityCouncilDistrictId.capital-projects.tsx
```diff
@@ -10,34 +10,37 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
   const page = pageParam === null ? 1 : parseInt(pageParam);
   const { cityCouncilDistrictId } = params;
   if (cityCouncilDistrictId === undefined || isNaN(page)) {
     throw json("Bad Request", { status: 400 });
   }
   const offset = (page - 1) * itemsPerPage;
 
-  const projectsByCityCouncilDistrictPromise =
-    findCapitalProjectsByCityCouncilId(
+  const projectstPromise = findCapitalProjects(
```
TangoYankee, line 21:
```suggestion
  const projectsPromise = findCapitalProjects(
```

app/routes/city-council-districts.$cityCouncilDistrictId.capital-projects.tsx
```diff
@@ -10,34 +10,37 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
 
   const agenciesPromise = findAgencies({
     baseURL: `${import.meta.env.VITE_ZONING_API_URL}/api`,
   });
 
-  const [agenciesResponse, projectsByCityCouncilDistrictResponse] =
-    await Promise.all([agenciesPromise, projectsByCityCouncilDistrictPromise]);
+  const [agenciesResponse, capitalProjectsResponse] = await Promise.all([
+    agenciesPromise,
+    projectstPromise,
```
TangoYankee, line 39:
```suggestion
    projectsPromise,
```

app/root.tsx
```diff
@@ -92,6 +99,8 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
         boroughs,
         communityDistricts: null,
         cityCouncilDistricts: null,
+        agencies,
+        // ENV,
```
TangoYankee, line 103:
```suggestion
```

app/components/AdminDropdown/index.tsx
```diff
@@ -40,6 +41,8 @@ export function AdminDropdown({
           )
         }
         value={selectValue ?? ""}
+        //TODO: add padding to text area to fix CloseIcon covering the ellipsis
```
TangoYankee, line 44: Create a ticket

```suggestion
```

app/components/AdminDropdown/AgencyDropdown.tsx
```diff
@@ -0,0 +1,46 @@
...
+    analytics({
+      category: "Dropdown Menu",
+      action: "Change Managing Agency",
+      name: nextManagingAgencyAcronym,
+    });
+
+    onSelectValueChange(nextManagingAgencyAcronym);
+  };
+
+  const agencyOptions = agencies
```
TangoYankee, line 28: I'll make tickets so that we're sorting this in the API and the frontend can serve them as-is

app/root.tsx
```diff
@@ -28,22 +29,21 @@ import {
...
   DistrictTypeDropdown,
   CommunityDistrictDropdown,
   CityCouncilDistrictDropdown,
+  AgencyDropdown,
 } from "./components/AdminDropdown";
-import {
-  GoToCityCouncilDistrictBtn,
-  GoToDistrictBtn,
-} from "./components/GoToDistrictBtn";
-import { GoToCommunityDistrictBtn } from "./components/GoToDistrictBtn/GoToCommunityDistrictBtn";
```
TangoYankee, line 44: If we're not using these components, we should clean up their definition files as well

TangoYankee:
Bones look good. A few suggestions on places we can tidy up. We should also clean up the commit history so that the code is functional at each commit (simplest approach is an interactive rebase where every commit after the first is a "fixup" into the first commit. It's also a chance to reword the first commit message to summarize the whole feature.)

TylerMatteo:
@TangoYankee I pushed a few commits for the clean up you suggested. I agree we should clean up the commit history. What do you think about just squashing into one commit? There really isn't a meaningful functional commit in the middle of it. I would of course make sure @dhochbaum-dcp is coauthored.

PR #153, opened 2025-02-07
144 Adblock Breaks Features

### LINKED ISSUE #152: Adblock Breaks Features
Errors caused by adblockers are breaking site functionality

dhochbaum-dcp:
The issue was caused by ad blocking extensions preventing Matomo from loading, causing the analytics function to fail.  This adds a simple check for window._paq to make sure if adblock is preventing Matomo from loading, it doesn't break features.

Closes #152

### DIFF
--- a/app/utils/analytics.js
+++ b/app/utils/analytics.js
@@ -1,19 +1,21 @@
 export function analytics(eventData) {
-  if (eventData.value) {
-    window._paq.push([
-      "trackEvent",
-      eventData.category,
-      eventData.action,
-      eventData.name,
-      eventData.value,
-    ]);
-  } else {
-    window._paq.push([
-      "trackEvent",
-      eventData.category,
-      eventData.action,
-      eventData.name,
-    ]);
+  if(window._paq) {
+    if (eventData.value) {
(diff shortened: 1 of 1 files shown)


app/utils/analytics.js
```diff
@@ -1,19 +1,21 @@
...
-      eventData.value,
-    ]);
-  } else {
-    window._paq.push([
-      "trackEvent",
-      eventData.category,
-      eventData.action,
-      eventData.name,
-    ]);
+  if(window._paq) {
```
TylerMatteo, line 2: what is `_paq`?

dhochbaum-dcp: It is what Matomo uses to track data.

TylerMatteo: Gotcha. Can you just add a comment to that effect?

TylerMatteo (changes requested):

app/utils/analytics.js
dhochbaum-dcp: Happy to.  It's been added.

TylerMatteo (approved):

Issue #180, opened 2025-05-19
Add total projects count to UI


TylerMatteo:
This issue is for adding the total projects count to the UI, made available in the API via https://github.com/NYCPlanning/ae-zoning-api/issues/413.

# Acceptance Criteria

- [ ] Add total projects count to the project list heading ([designs](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=3067-13454&m=dev))
- [ ] Add total projects range count to project list below pagination ([designs](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=4308-44267&m=dev)).

PR #182, opened 2025-05-19
178 Design Feedback: No results message

### LINKED ISSUE #178: Design Feedback: No results message
Not sure if the phrasing is the best as it might indicate there were projects that do exist for the options I selected, but I have reached the end of project list stack.

Image

edit suggestion below:

Image

dhochbaum-dcp:
Added a message when the combination of applied filters returns 0 results:

image

Closes #178


TylerMatteo:
@dhochbaum-dcp this seems to include changes for both #178 and #180. Please break it up.

app/components/CapitalProjectsList/CapitalProjectsList.tsx
```diff
@@ -2,18 +2,25 @@ import { Box, Collapse, Flex, Text, VStack } from "@nycplanning/streetscape";
...
 }
 
 export const CapitalProjectsList = ({
   capitalProjects,
   agencies,
+  capitalProjectsTotal,
   isExpanded,
 }: CapitalProjectsListProps) => {
+  if (capitalProjectsTotal === 0) {
+    return <CapitalProjectsListNoResults />;
```
TylerMatteo, line 21: I don't really think this warrants being it's own component. You can put the contents of that component directly in this `if`.

TylerMatteo (changes requested):

app/components/CapitalProjectsList/CapitalProjectsList.tsx
dhochbaum-dcp: Done.

TylerMatteo (approved):

PR #195, opened 2025-07-17
React Router 7 Migration

TylerMatteo:
# Summary

This PR migrates the app from Remix v2 to React Router v7. As [React Router 7 is the next version of Remix](https://remix.run/blog/react-router-v7), you can think of this as a major upgrade as much as a migration. We followed the instruction [here](https://reactrouter.com/upgrading/remix) and used the codemod linked there for as much of the changes as possible. Summarizing the steps:
* Adopt all [Remix future flags](https://remix.run/docs/en/main/start/future-flags)
* Upgrade to most recent Remix dependencies
* Switched routing to config-based rather than the existing file based. `routes.ts` now configures the routing for the app, rather than the names of files in `/routes`. I intentionally left the file names and structure in that folder as-is with the intention that we would refactor and reorganize them as a separate PR.
* Ran the code mod, switched over dependencies to corresponding React Router deps
* Deleted `entry.client.tsx` and `entry.server.tsx`. These files are optional in React Router and leaving them as-is was causing rendering issues when trying to run the app.


app/components/AdminDropdown/index.tsx
```diff
@@ -1,9 +1,4 @@
-import {
-  FormControl,
-  FormLabel,
-  IconButton,
```
horatiorosa, line 4: Do we not need the `IconButton` anymore?

horatiorosa (approved):
seems ok

TangoYankee (commented):
The last half of commits are mine. I don't think I can be an objective reviewer.

PR #215, opened 2025-10-04
Layout, state management, and routing rework

TylerMatteo:
# Intro
This PR contains a large set of changes done by myself and @horatiorosa that are intended to accomplish the following goals while leaving the application in a functional, deploy-able state where all functionality of the app still works as expected:

* Overhaul how the responsive layout of the app works to align with designs for the overall Community Board Budget Requests project by introducing a responsive grid system
* Add the new "Welcome Panel" shown in [CBBR designs](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=5645-92270&m=dev), including "Don't show this again" functionality.
* Rethink how the related systems of state management and routing (including query parameters) work throughout the app with the ultimate goal of having Capital Project filters take effect immediately on change instead of via the "Search" button that exists in production right now
* While tackling the above goals, take opportunities to refactor code where appropriate

## Things not covered in this change

* Updates to UI for [listing capital projects](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=4601-51492&m=dev) and individual capital project details that will be done as part of the broader CBBR project are intentionally not included in this PR. Note that this includes moving the project "attribute" filters from the left-hand panel to the right-hand panel under the "filters" button. The "Search By Attribute" filters have instead been left in the left-hand panel and had their styling updated to fit into new designs.
* Any references to Community Board Budget Requests. These changes is meant to set up the app for adding CBBR functionality while leaving it in a deploy-able state without half-implemented features.

## Grid system
_TODO_

## Routing and State Management for Capital Project filters

Because routing and state management are so closely related in this app, its easiest to talk about these changes together. Overall, the goals for these changes are:
* Have filters take effect immediately on change, so long as the new input the user is attempting to set is valid.
* A user should be able to share a link to the app as they're seeing it and expect someone opening that link to see the same filters applied
* To centralize state management, rely on query parameters where possible, as opposed to `setState`

### How does it work before these changes?
To set a baseline for these changes, I'll summarize how this work in the app on `main`.

#### Admin boundary filters
image

On `main`, state for these filters is managed entirely through routing. The `districtType`, `boroughId`, and `districtId` query parameters always track the value in the corresponding dropdown. However, the _filters actively being applied to projects_ are captured in path params. The 3 routes we have that show a list of capital projects are:
* `/capital-projects` - this route means no admin filters are being applied
* `/boroughs/:boroughId/community-districts/:communityDistrictId/capital-projects` - this is the route when filtering by a Community District
* `/city-council-districts/:cityCouncilDistrictId/capital-projects` - this it the route when filtering by a City Council District

The version of the app in `main` needed to use both path and query params because the value in a given dropdown can be different from the filters actually being applied. On when a user clicks "Search" do we "sync" up the values and apply new filters.

#### Attribute filters
image

On `main`, state for these filters is managed through a combination of query params and React `useState`s declared in `root.tsx`. The `useState` variables track the values as the appear in each input. In contrast to admin filters, query params track the _filters actively being applied to projects_. When a user clicks "Search" the values in the React state are "synced" to the query params.

### How does it work with these changes?

Now that we want filters to take effect immediately upon change, we have an opportunity to refactor a lot of code. Needing to keep track of the current input values separately from the filters actually being applied introduced some complexity that we no longer need. To accomplish this, I took the approach of using query parameters for state management wherever possible and only using other tools like `useState` where necessary. The query parameters that control these filters (`districtType`,`boroughId`,`districtId`,`managingAgency`,`agencyBudget`,`commitmentsTotalMin`, and `commitmentsTotalMax`) are simply updated when a new valid input is given by the user.

#### Moving code from `root.tsx` to lower level components

`root.tsx` has gotten quite long and unwieldy on `main`. This has been hard to avoid up until now because moving state and logic to lower components would have required a lot of prop drilling or setting up context providers. Now that we can rely more on query parameters for state management, we can treat that as a kind of context provider.

Two components used in `root.tsx`, `FilterMenu` and `SearchByAttributeMenu`, were mostly presentational. The component for the individual filter inputs were used in `root` and passed as children to these components.  Now that it's easier to access state lower in the component tree, I decided to move most of this code into these components. `root.tsx` passes loader data like `agencies` and `communityDistricts` to these components and they handle the input components from there. Note that now neither of those components take `children`. The individual input components like `AgencyDropdown` are mostly unchanged.

#### `useUpdateSearchParams`

I added this new hook to make updated a query parameter _without changing any of the existing ones_ easier. You no longer have to spread in the object of existing params then set your new one like you'd have to do with `setSearchParams`. The hook signature is meant to build on `setSearchParams`'s by being an array with one more value for the new function.

#### Fewer pages in `/routes`

Now that the selected district is just tracked by query params, we don't need these two routes:
* `/boroughs/:boroughId/community-districts/:communityDistrictId/capital-projects`
* `/city-council-districts/:cityCouncilDistrictId/capital-projects`

I removed the files for those routes and their entries in `routes.ts`. The changes to `capital-projects.tsx` make that page work with the filters if they are set, matching the functionality from the deleted pages.

#### `ProjectAmountMenu`

The project amount filter still uses `useState`. This is because we only want to the applied filter to update when the user's input is _valid_. If they type in a minimum larger than their maximum, the input should still change and they see the error message, but the actual filter assigned to query param shouldn't change. I did this by moving the code for this component from `root.tsx` into `ProjectAmountMenu/index.tsx` and adapting it to the new requirements.


horatiorosa:
> * "Welcome" text should be centered vertically in the Accordion
> * Main text area of the box is missing horizontal padding
> * The height and spacing of all of the items at the bottom is incorrect
> 
> There are similar issues in the Capital Projects list, but I assume those will be addressed in #217.
> 
> There are also notable inconsistencies with the font sizing - ex. the font size of the Accordion headers is different on the left panel from the size on the right panel, where in the designs they are equal.

Yeah, you're right. I actually addressed this (and a number of other CSS fixes) in a separate pr but I haven't yet pulled them into here. Good to note here so it doesn't get lost in the flux.

TangoYankee:
I think "resetting selection" should include clearing the state of the amount selector. 

[reset-selection-project-amount.webm](https://github.com/user-attachments/assets/689d609b-1031-46df-8fa6-992c74f963cd)

TangoYankee:
The "Reset Selections" looks like it's only the width of the text. However, the button goes the full width of the panel. I got confused because I would think I was clicking away to activate the selection, when I was actually clicking on "Reset Selections"

dhochbaum-dcp:
I'd assume it would be more appropriate to address in #217 rather than this pull request, but I want to note: the "Managing Agency" select has 170 choices, most of which return no results, is this intended?

dhochbaum-dcp:
"Capital Projects Portal" text does not appear in the header bar, regardless of screen width.

TangoYankee:
> I'd assume it would be more appropriate to address in #217 rather than this pull request, but I want to note: the "Managing Agency" select has 170 choices, most of which return no results, is this intended?

I meant to bring this up in sprint planning but it fell through the cracks. I made a draft PR where we add a `capital-projects/managing-agencies` endpoint, akin to what we do for budget requests. The intent would be to switch the capital projects agency dropdown to that endpoint, instead of the "agencies" endpoint it currently uses.

https://github.com/NYCPlanning/ae-zoning-api/pull/509

TylerMatteo:
> I think "resetting selection" should include clearing the state of the amount selector.


Fixed.

TylerMatteo:
> Is there documentation of how the automatic zoom is intended to work?
> 
> Currently, only making a selection of Community District zooms to an area. Selecting a City Council District or any of the attributes does not change the zoom.
> 
> Having selected a Community District, clearing any/all of the Location filters does not update the zoom. No matter what, "Reset Selections" zooms out to default.

[Fixed](https://github.com/NYCPlanning/ae-cp-map/pull/215/commits/5a4985d77c56070bde1cb4a6ba7a8e2146650ee5) the zooming to City Council Districts. I agree there's probably room for improvement for other zoom out use cases but maybe that's a good candidate to discuss with design and handle after this PR?

dhochbaum-dcp:
> > I'm getting some 404 errors when using the right pane's "back" button:
> 
> @dhochbaum-dcp I just pushed a commit that should fix this, can you double check?

The good news: I'm not getting 404 errors.
The bad news: It doesn't work how I'd expect, which is to retain the currently applied filters but change the path to return the user to the results list.  Right now it appears to just duplicate the functionality of the browser's "back" button.

PR #231, opened 2025-10-25
Add zoom and bearing reset component to new grid layout

### LINKED ISSUE #201: Add zoom and bearing reset component to new grid layout
- [x] Add custom buttons for zoom and bearing reset per [designs](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=5468-43643&m=dev)
- [x] Button click behavior should match built in Deck [zoom](https://deck.gl/docs/api-reference/widgets/zoom-widget) and [compass](https://deck.gl/docs/api-reference/widgets/compass-widget) widgets in current production version of the app.

horatiorosa:
#### Description
Create ZoomPanel component for the zoom and bearing controls and add to root
Remove default deck gl controls

#### Tickets
Closes #201

### DIFF
--- a/app/components/MapViewControls.tsx
+++ b/app/components/MapViewControls.tsx
@@ -1,0 +1,144 @@
+import { VStack } from "@chakra-ui/react";
+import {
+  Tooltip,
+  ButtonGroup,
+  Icon,
+  IconButton,
+  CompassIcon,
+  AddIcon,
+  MinusIcon,
+} from "@nycplanning/streetscape";
+import { LinearInterpolator } from "@deck.gl/core";
+import type { MapViewState } from "@deck.gl/core";
+
+type mapViewControlsProps = {
+  viewState: MapViewState;
+  setViewState: (next: MapViewState) => void;
+  step?: number;
+  minZoom?: number;
+  maxZoom?: number;
+};
+
+export function MapViewControls({
+  viewState,
+  setViewState,
+  step = 1,
+  minZoom = 10,
+  maxZoom = 20,
+}: mapViewControlsProps) {
+  const clamp = (value: number) => Math.min(maxZoom, Math.max(minZoom, value));
+
+  const animate = (partial: Partial<MapViewState>) =>
+    setViewState({
+      ...viewState,
+      ...partial,
+      transitionDuration: 500,
+      transitionInterpolator: new LinearInterpolator(["zoom", "bearing"]),
+    });
+
+  const zoomBy = (delta: number) =>
+    animate({ zoom: clamp(viewState.zoom + delta) });
+  const resetCompass = () => animate({ bearing: 0 });
+
+  return (
+    <VStack rowGap={"0"}>
+      <ButtonGroup isAttached={true} orientation="vertical" spacing={0}>
+        <IconButton
+          className="zoomin"
+          variant={"secondary"}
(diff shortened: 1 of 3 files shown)


app/components/ZoomPanel.tsx
```diff
@@ -0,0 +1,113 @@
...
+            size={"sm"}
+            icon={<Icon as={AddIcon} boxSize={4} color={"gray.600"} />}
+            borderRadius={0}
+            borderTopLeftRadius={4}
+            borderTopRightRadius={4}
+            _hover={{
+              backgroundColor: "white",
+              border: "none",
+            }}
+            _focus={{
```
TylerMatteo: Please check in with Natasha about a focus styling for these buttons. I don't mind not having styling on hover but no focus styling is an accessibility concern.

TylerMatteo (changes requested):

app/components/ZoomPanel.tsx
```diff
@@ -0,0 +1,180 @@
...
+    setViewState({
+      ...viewState,
+      ...partial,
+      transitionDuration: 500,
+      transitionInterpolator: new LinearInterpolator(["zoom", "bearing"]),
+    });
+
+  const zoomBy = (delta: number) =>
+    animate({ zoom: clamp(viewState.zoom + delta) });
+  const resetMap = () => animate({ bearing: 0 });
```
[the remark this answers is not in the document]
horatiorosa: https://github.com/user-attachments/assets/302d15cd-05eb-495c-a1d1-e6a1a89a98cf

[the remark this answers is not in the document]
pratishta: Awesome, thanks! Resetting the map makes me think that all the params should be what they were when we load the page. My suggestion would be, since we have `INITIAL_STATE_VIEW` to use it to set the default lat, long, and zoom values as well as the bearing and pitch. 

`const resetMap = () => animate(INITIAL_VIEW_STATE);`

If we don't want that feature / only care about resetting the compass, I'd recommend changing the name to `resetCompass`

app/components/ZoomPanel.tsx
```diff
@@ -0,0 +1,180 @@
...
+    });
+
+  const zoomBy = (delta: number) =>
+    animate({ zoom: clamp(viewState.zoom + delta) });
+  const resetCompass = () => animate({ bearing: 0 });
+
+  return (
+    <VStack rowGap={"0"}>
+      <ButtonGroup isAttached={true} orientation="vertical" spacing={0}>
+        <Tooltip
```
TylerMatteo: These tooltips don't seem to ever close once opened on mobile with touch events. Since you can't have hover effects on touch devices, I think she should just never show up on them?

TylerMatteo (changes requested):

app/components/ZoomPanel.tsx
horatiorosa: There seems to be inconsistent behavior depending on the browser:
 - FF mobile emulator, the tooltips remain
 - Chrome and Safari, they do close after clicks/touches.

Should we get rid of the tooltips altogether @TylerMatteo?

horatiorosa: I will remove the tooltips. It doesn't seem to work on mobile as I'd hoped it would.

TylerMatteo (approved):

PR #246, opened 2025-11-05
Use variables defined from environment

### LINKED ISSUE #237: Use variables defined from environment
## Acceptance criteria

- [ ] Replace use of zoning api url environment variable with variable defined in `envFlags`
- [ ] Define basemap server variable in `envFlag` and replace use of environment variable with it
- [ ] Define cpdb data url variable in `envFlag` and replace use of environment variable with it
- [ ] Throw an error when an environment variable is not set
  - One approach could be a function that returns an object of validated environment variables

horatiorosa:
#### Description
Replace various API calls with variables defined in `envFlags` to minimize duplication and for readability.

#### Tickets
Closes #237


app/utils/env.ts
```diff
@@ -0,0 +1,21 @@
+const envFlags = import.meta.env;
+
+function validateEnv(envFlags: ImportMetaEnv) {
+  const missing: string[] = [];
+
+  if (!envFlags.VITE_ZONING_API_URL) missing.push("VITE_ZONING_API_URL");
+  if (!envFlags.VITE_BASEMAP_URL) missing.push("VITE_BASEMAP_URL");
+  if (!envFlags.VITE_CPDB_DATA_URL) missing.push("VITE_CPDB_DATA_URL");
+
+  if (missing.length) {
```
TangoYankee:
```suggestion
  if (missing.length > 0) {
```

TangoYankee (approved):
🚀 🌔

Issue #255, opened 2025-11-12
Individual CBBR panels getting cut off at varying screen sizes


pratishta:
When navigating to an individual CBBR and adjusting the height of the desktop screen size, the panel gets cut off and there's no scroll to see all of it. 
<img width="1710" height="703" alt="Image" src="https://github.com/user-attachments/assets/87859f81-91a0-4feb-b68b-a5bd8979840e" />

Issue #245, opened 2025-11-03
Select CBBR on map zooms and navigates to individual CBBR view


TylerMatteo:
## Acceptance Criteria

- [ ] Clicking on a single CBBR on the map should navigate to the page for that CBBR
- [ ] When a CBBR is selected, map should "fly to" its location, similar to existing behavior for Capital Projects

PR #277, opened 2025-11-26
245/individual cbbr panel

pratishta:
Closes https://github.com/NYCPlanning/ae-cp-map/issues/245
Closes https://github.com/NYCPlanning/ae-cp-map/issues/255

Notes
- This now uses the pre-packing icon strategy to help cause less conflicts with the clustering work
- I added the ~hover and~ selected behavior for the icon markers


pratishta:
Ready for review again since hovering is out of scope for this ticket >:)

Issue #259, opened 2025-11-17
Add hover effects for CBBR and Capital Project lists


TylerMatteo:
## Acceptance Criteria

- [ ] When a capital project or budget request is hovered in their respective lists, background color should chance from `gray.50` to `gray.100` ([designs](https://www.figma.com/design/LYHHoPop9l0jpEivk5CFzJ/Capital-Projects-Portal?node-id=4602-53749&m=dev))
- [ ] When a capital project or budget request is hovered in their respective lists, the corresponding project or request on the map is highlighted with its hover effect

PR #301, opened 2025-12-11
259/add hover effects

pratishta:
Closes https://github.com/NYCPlanning/ae-cp-map/issues/259

~Additionally, this PR also filters out `community-board-budget-request-label` data so that we don't get icon markers for CBBRs with polygons.~

### DIFF
--- a/app/components/layers/useCapitalProjectsLayer.client.tsx
+++ b/app/components/layers/useCapitalProjectsLayer.client.tsx
@@ -28,8 +28,14 @@ export interface CapitalProjectProperties {
   commitmentsTotal: number;
 }
 
-export function useCapitalProjectsLayer(opts?: { visible?: boolean }) {
-  const visible = opts?.visible ?? true;
+export function useCapitalProjectsLayer(opts: {
+  visible?: boolean;
+  hoveredCapitalProject?: string | null;
+  setHoveredOverProject: (newHoveredOverProject: string | null) => void;
+}) {
+  const visible = opts.visible ?? true;
+  const hoveredCapitalProject = opts.hoveredCapitalProject;
+  const setHoveredOverProject = opts.setHoveredOverProject;
   const { managingCode, capitalProjectId } = useParams();
   const [searchParams] = useSearchParams();
   const managingAgency = searchParams.get("managingAgency");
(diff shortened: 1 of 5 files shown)


TylerMatteo (approved):

app/components/layers/useCapitalProjectsLayer.client.tsx
```diff
@@ -108,6 +115,15 @@ export function useCapitalProjectsLayer(opts?: { visible?: boolean }) {
     getPointRadius: 5,
     getLineColor: [255, 255, 255, 255],
     getLineWidth: 1,
+    onHover: (data) => {
+      const managingCodeCapitalProjectId =
+        data.object?.properties?.managingCodeCapitalProjectId;
+      if (data.index === -1) {
+        setHoveredOverProject(null);
+      } else if (managingCodeCapitalProjectId) {
```
TangoYankee: Truthy check will fail when value is 0, empty string, or some other falsy value. Prefer explicit checks, like a negation of undefined/null. 

```
managingCodeCapitalProjectId !== undefined
```

app/components/layers/useCommunityBoardBudgetRequestsLayer.client.tsx
```diff
@@ -158,11 +170,19 @@ export function useCommunityBoardBudgetRequestsLayer(opts?: {
...
+    onHover: (data) => {
+      const id = data.object?.properties?.id;
+      if (data.index === -1) {
+        setHoveredOverCbbr(null);
+      } else if (id) {
+        setHoveredOverCbbr(id);
+      }
+    },
+    onClick: (data) => {
+      setClickInfo({ id: data.object?.properties?.id, clicked: data.picked });
```
TangoYankee: Does the `picked` value toggle between `true` and `false`? From what I can tell,  we always check whether the project id matches the clicked id, and we check whether its "picked/clicked" status was true. We can simplify these other checks by resetting the 'clicked id` to "nothing" when "picked" is false.

this value reset becomes:

```ts
const id = data.object?.properties?.id;
const nextClickedId = data.picked && id !== undefined ? id : null;
setClickInfo(nextClickedId);
``` 

Refactoring the state:
(NULL is a more idiomatic expression of "no value" than an empty string)

```ts
[clickedId, setClickedId] = useState<string | null>(null)
```

The checks become:

```ts
      if (clickedId === id) {
```

Issue #308, opened 2025-12-17
Refine checking for valid values in icons and hover


pratishta:
### Acceptance criteria

- [ ] Use explicit values when checking for valid ids on hovering ([comment](https://github.com/NYCPlanning/ae-cp-map/pull/301#discussion_r2620486488))
- [ ] Use fall back when assigning policy area based svgs 

<img width="1697" height="57" alt="Image" src="https://github.com/user-attachments/assets/d84df302-5197-4095-97bd-0906225f7f6e" />

PR #313, opened 2025-12-18
Update conditionals and remove unnecessary icon mapping in geojson layer

pratishta:
Closes https://github.com/NYCPlanning/ae-cp-map/issues/308

In removing the "undefined.svg" error, I moved all the conditional click behavior to the geojson layer since that's where we should be seeing the clicked behavior.

### DIFF
--- a/app/components/layers/useCommunityBoardBudgetRequestsGeoJsonLayer.client.tsx
+++ b/app/components/layers/useCommunityBoardBudgetRequestsGeoJsonLayer.client.tsx
@@ -3,9 +3,23 @@ import { CommunityBoardBudgetRequestGeoJson } from "~/gen";
 import { useParams } from "react-router";
 import { FlyToGeoJsonExtension } from "../../extensions";
 import { env } from "~/utils/env";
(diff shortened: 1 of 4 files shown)


app/components/layers/useCommunityBoardBudgetRequestsGeoJsonLayer.client.tsx
```diff
@@ -29,13 +44,14 @@ export function useCommunityBoardBudgetRequestsGeoJsonLayer() {
...
     pointType: "icon",
     getIcon: (d: { properties: CommunityBoardBudgetRequestProperties }) => {
-      const icon = policyAreaIconsMap[d.properties.policyAreaId];
-      return {
-        url: `/policy-area-icons/${icon}.svg`,
-        width: 40,
-        height: 40,
-      };
+      const icon = policyAreaIconsMap[d.properties?.cbbrPolicyAreaId];
+      return `${icon}-click`;
```
TangoYankee: What happens if/when "properties" is `undefined`?

TangoYankee: Or is it reliably always defined and we can remove the "?"

Issue #314, opened 2025-12-23
[Bug: Priority]: CBBR Icons are misaligned


mbh329:
### What happened? What did you expect to happen?

CBBR icons shown when filtered don't appear to line up with the correct Policy Area. In the screenshot below I've filtered for Healthcare and Human Services as the Policy Area, and having selected a CBBR you can see that it's in the correct Policy Area, but the icons shown are the icons for Youth Education and Child Welfare — I've clicked on and off some other Policy Area filters and they similarly are showing mismatched.

<img width="1919" height="914" alt="Image" src="https://github.com/user-attachments/assets/ea2f7cb2-f3a0-4a20-b827-3deadcc88f5b" />

### What browsers are you seeing the problem on?

Chrome

### Which operating system are you using?

Windows

TangoYankee:
I imagine this is a misalignment between the policy area ids as they're listed in the frontend v the api. We did reorder the policy area ids in the middle of the development process. It's possible the front end has the correct ids but the database needs to be updated by rerunning data flow.

It's also possible the front end has the incorrect the ids and needs to be updated to match the database/data flow.

**Note to investigating developer:** before troubleshooting locally, ensure that your local database is a fresh run of the cbbr data flow

PR #316, opened 2026-01-06
Adjust policy area icon ids

pratishta:
Closes https://github.com/NYCPlanning/ae-cp-map/issues/314


TangoYankee (commented):
I'd like to look into this more. Based on the source file "cbbr_options_no_duplicates_surrogate_ids_v2", the mapping between the ID and the Policy Area should already be correct

TangoYankee:
> Also not sure if this is related, but icons for "Other needs" do not show up when selected in the dropdown filter for any instances of CP Map.

I noticed that, as well. It might be a quirk of the data where every request that falls within "Other Needs" is also unmapped. It does seem worth looking into.

Issue #327, opened 2026-01-26
Make entire Map Layers Panel collapsable


TylerMatteo:
## Acceptance Criteria

- [ ] Entire Map Layers Panel expands and collapses when user click "Capital Planning Data" header ([designs](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=107-34398&m=dev))
- [ ] Filter By Location and How to Use This Tool accordions still exist within panel. Note smaller font size for these headings
- [ ] All changes should be behind `facDbPhase1` feature flag.

pratishta:
@TylerMatteo Yes. I just updated it

PR #351, opened 2026-02-04
Make capital planning data one collapsible item

pratishta:
Closes https://github.com/NYCPlanning/ae-cp-map/issues/327


app/components/SearchByCbbrMenu.tsx
```diff
@@ -78,7 +79,7 @@ export const SearchByCbbrMenu = ({
       borderRadius={"sm"}
       borderWidth={"1px"}
       marginTop={2}
-      marginX={2}
+      marginX={env.facDbPhase1 == "ON" ? "" : "2"}
```
TylerMatteo: Can this be `0 : 2` instead of `"ON" ? "" : "2"`? The number 2 (as opposed to string `"2"`) points to size token in Streetscape.

TylerMatteo (approved):

PR #353, opened 2026-02-06
319 Add cross hatch symbology for Capital Projects

### LINKED ISSUE #319: Add cross hatch symbology for Capital Projects
## Acceptance Criteria
- [x] Update the styling for mapped capital projects to include a cross hatch pattern [per designs](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=622-28087&m=dev). Styling for point projects is unchanged
- [ ] Update Capital Project legend icon next to layer toggle in map layers panel
- [x] All changes should be behind `facDbPhase1` feature flag.

## Notes
* Reach out to @Ntoal if you need any assets exported from Figma
* Adding cross hatch pattern in Deck will likely be done with [FillStyleExtension](https://deck.gl/docs/api-reference/extensions/fill-style-extension)

dhochbaum-dcp:
This PR:
- Updates the legend icons for Capital Projects and CBBRs
- Creates the new icon to be used for Facilities
- Updates the Captial Projects layer to use a cross-hatch matching the designs

The technical solution requires building from the [FillStyleExtension](https://deck.gl/docs/api-reference/extensions/fill-style-extension).  All attempts to use the extension in the documented fashion failed, as there is no way to get rid of pixelation that occurs at high zoom levels.  Instead, I created two similar extensions of the FillStyleExtension, one for polygons and one for points.

Both extensions fill the layer with the repeating pattern of a single light orange pixel, as an image is required for the FillStyleExtension.  They then inject a custom function into the the shaders which fill the area with the desired lines.

The use of two separate extensions is required because the sublayers need to be targeted separately, otherwise the lines will also appear over the white borders of points.  Even with splitting out the sublayers between `polygons-fill` and `points-circle`, there is some conflict as each will still by default write over the borders of points.  Because of this, the PolygonFill checks the geometry to make sure it is not overwriting the edge of a point, and the PointFill checks the geometry and writes both the pattern for the fill and the white border.

Closes #319


dhochbaum-dcp:
@Ntoal this has been pushed to QA, thought you might want to take another look so you can see it combined with the previous map changes.

@TylerMatteo I did not realize until merging the PR that switches the filtering from the backend to the frontend that the way the shaders were being implemented would break the method of filtering on the front end.  In order to make it work, I had to apply the `MaskExtension` and `DataFilterExtension` and all of their corresponding properties to both sublayers in addition to the layer.  I didn't notice any impact on performance but thought I'd call it out in case you had an idea of a better way to handle this.

Ntoal:
Approved - thank you!

TylerMatteo:
> @TylerMatteo I did not realize until merging the PR that switches the filtering from the backend to the frontend that the way the shaders were being implemented would break the method of filtering on the front end. In order to make it work, I had to apply the `MaskExtension` and `DataFilterExtension` and all of their corresponding properties to both sublayers in addition to the layer. I didn't notice any impact on performance but thought I'd call it out in case you had an idea of a better way to handle this.

Thanks for calling this out. Duplicating the code like this isn't ideal but we can't really spend any more time on this so I'm willing to take it on as some tech debt. We'll have to find some time down the line to refactor this implementation.

TylerMatteo (approved):

PR #388, opened 2026-02-24
376 New location filtering UI for Community Boards and City Council Districts

### LINKED ISSUE #376: New location filtering UI for Community Boards and City Council Districts
## Acceptance Criteria
- [ ] Implement [new district filtering UI](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=90-22649&p=f&m=dev) for existing Community Board and City Council District filtering
- [ ] By default, the map should open with Community Districts selected and displayed.
- [ ] After a location filter button is chosen, a CPP user must be able to select geographies directly on the map by clicking or tapping anywhere in the geography not covered by a CBBR or Capital Project
- [ ] Hovering over a geography must highlight that area to indicate what would be selected
- [ ] Users can only select one geography at a time. Clicking a new geography when one is already selected will replace the current selection
- [ ] Switching geography type when a geography is selected with clear the active geographic filter but app should still track the selected geography. If user switches back to previous geography type, the app should filter to the previously selected geography
- [ ] When a geography is selected
    - [ ] pan and zoom (fly to) the selected geography
    - [ ] visually highlight the selected boundary on the map
    - [ ] update results immediately across all CPP datasets
- [ ] users can de-select a geography by clicking or tapping it again.
- [ ] All changes are behind `facDbPhase1` feature flag

### Technical requirements
- [ ] Update how geospatial filters are applied from using different tile endpoints to front end masking using Deck [MaskExtension](https://deck.gl/docs/api-reference/extensions/mask-extension). See [PoC branch](https://github.com/NYCPlanning/ae-cp-map/tree/344/POC-masking) for guidance
- [ ] Update existing UI grid with new row for "stacking" geographic filter component on smaller screen sizes

dhochbaum-dcp:
- Implements [new district filtering UI](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=90-22649&p=f&m=dev) for existing Community Board and City Council District filtering
- Adds highlighting districts on hover, makes map districts selectable
- Changes map geo filtering to be performed on the Front-end using MaskExtension

image

This adds two new layers, which are necessary for the hover to work properly.  Without the new layers, the hover color for a district is frequently over-ridden by the standard color of an adjacent layer.  The new layers display the district boundaries, and sit beneath the existing layers, where the district boundaries have been made invisible unless the district is hovered.

Known issues:
- Cursor is pointer when hovering areas that are in Community Districts layer but are not Community Districts (ex. JFK)
- Issues with the border-radius of the filtering UI that will require a separate PR and update to Streetscape
   - https://github.com/NYCPlanning/ae-streetscape/pull/130

Closes #376

### DIFF
--- a/app/layouts/MapPage.tsx
+++ b/app/layouts/MapPage.tsx
@@ -3,7 +3,6 @@ import {
   GridItem,
   Accordion,
   Box,
-  Collapse,
   AccordionItem,
   AccordionButton,
   Heading,
@@ -12,6 +11,9 @@ import {
   Text,
   HStack,
   Link,
+  Tab,
+  TabList,
+  Tabs,
 } from "@nycplanning/streetscape";
 import {
   Outlet,
@@ -39,10 +41,10 @@ import {
   CommunityBoardBudgetRequestAgencyCategoryResponseId,
   CommunityBoardBudgetRequestNeedGroupId,
   CommunityBoardBudgetRequestPolicyAreaId,
+  DistrictId,
   DistrictType,
 } from "../utils/types";
 import { HowToUseThisTool } from "../components/AdminDropdownContent/HowToUseThisTool";
-import { MapLayersPanel } from "../components/AdminMapLayersPanel";
 import {
   CapitalProjectLayerToggle,
   CommunityBoardBudgetRequestLayerToggle,
@@ -53,6 +55,7 @@ import type { RootContextType } from "../root";
 import { MapViewControls } from "~/components/MapViewControls";
 import { SearchByCbbrMenu } from "~/components/SearchByCbbrMenu";
 import { useState } from "react";
+import { MapLayersPanel } from "~/components/AdminMapLayersPanel";
 
 export const loader = async ({ request }: LoaderFunctionArgs) => {
   const url = new URL(request.url);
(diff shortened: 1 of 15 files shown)


app/layouts/MapPage.tsx
```diff
@@ -219,6 +233,306 @@ export default function MapPage() {
...
+  const getDefaultIndex = (type: DistrictType | null) => {
+    switch (type) {
+      case "ccd":
+        return 1;
+      default:
+        return 0;
+    }
+  };
+
+  const [savedGeoSelection, setSavedGeoSelection] = useState<{
```
TylerMatteo: Instead of creating this new state we have to keep track of, lets have separate query parameters for community boards versus city council districts. That way we can continue using query params as a single source of truth. `districtyType` can still track which layer should be on and masked to but the respective district Id params can be kept in the URL so they're "remembered" when the user switches back and forth.

TylerMatteo: Instead of creating this new state we have to keep track of, lets have separate query parameters for community board versus city council district Ids. That way we can continue using query params as a single source of truth. `districtyType` can still track which layer should be on and masked to but the respective district Id params can be kept in the URL so they're "remembered" when the user switches back and forth. I acknowledge changing `districtId` has broader implications but I'm comfortable taking the time to address it now. It will likely be easier to update code that might be behind the feature flag to use the new query params (such as the old district filter dropdowns) to use the new query params than it would be to have both sets of query params exist in the code at the same time.

TylerMatteo (changes requested):

app/layouts/MapPage.tsx
dhochbaum-dcp: A couple thoughts:
a) I think even after switching to separate query parameters, it makes more sense to keep this info in the state and not in the URL.  The URL shouldn't reflect your history; it makes the URLs unnecessarily long; sharing URLs with the history could lead to confusion and/or unwanted sharing of data.

b) It makes more sense to me to implement changing the query parameters when updating to make multiple selections, rather than in this.

TylerMatteo (dismissed):

Ntoal:
This looks great! Thank you!

horatiorosa (approved):
Visually matches the designs, hover and select work as expected, no code smells either

PR #410, opened 2026-03-20
Task/403 add search input to UI

### LINKED ISSUE #403: Add Search Input to UI
### Summary
This will be the first of several issues to add address and radius filtering to CPP. The scope of this Issue will implement the feature in an incomplete state but will hide the changes behind a feature flag

#### Acceptance criteria

- [x] Add new input to header bar using existing `Combobox` component in Streetscape according to [designs](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=899-52143&m=dev) 
- [x] Sample data should be hardcoded as is done in Streetscape
   - [x] The sample data should match the first 5 responses for a search for `123` in [Geosearch](https://geosearch.planninglabs.nyc/docs/)
   - [x] Coordinate data should be included so that the map pin issue can be worked on independently after
- [x] All changes should be behind existing Facdb phase 1 feature flag

#### Out of scope
-  Connecting the input to the Geosearch API and adding the pin to the map are separate issues
- The "Filter by Radius" button and popover are out of scope and will added following completion of this Issue.

horatiorosa:
#### Description
Issue 403: Add Search Address Input to the UI
 - refactored header from flexbox to css grid to take advantage of subgrid
 - added Combobox component to Streetscape
 - added array of test addresses to simulate search
 - added new `facDbPhase2` env flag

### Tickets
Closes #403

### DIFF
--- a/app/components/HeaderBar/AddressSearch.tsx
+++ b/app/components/HeaderBar/AddressSearch.tsx
@@ -1,0 +1,300 @@
+import {
+  Combobox,
+  useListCollection,
+  MapPinIcon,
+  Text,
+} from "@nycplanning/streetscape";
+import { SearchIcon, CloseIcon } from "@chakra-ui/icons";
+import { useFilter } from "@ark-ui/react/locale";
+import {
+  ComboboxInputValueChangeDetails,
+  ListCollection,
+} from "@ark-ui/react/combobox";
+import type { ComboboxCollectionItemProps } from "@nycplanning/streetscape";
+
+const geoSearchResults = [
+  {
+    type: "Feature",
+    geometry: {
+      type: "Point",
+      coordinates: [-73.958537, 40.809706],
+    },
+    properties: {
+      id: "288546",
+      gid: "nycpad:venue:288546",
+      layer: "venue",
+      source: "nycpad",
+      source_id: "288546",
+      country_code: "US",
+      name: "1231 AMSTERDAM AVENUE",
+      housenumber: "1231",
+      street: "AMSTERDAM AVENUE",
+      postalcode: "10027",
+      accuracy: "point",
+      country: "United States",
+      country_gid: "whosonfirst:country:85633793",
+      country_a: "USA",
+      region: "New York",
+      region_gid: "whosonfirst:region:85688543",
+      region_a: "NY",
+      county: "New York County",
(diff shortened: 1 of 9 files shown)


dhochbaum-dcp:
I took a look at the mobile view and noticed some issues: 
image

- "About" is missing padding on the right.
- The search bar should be in a second row
- The width of the drop-down appears to be the same as the search bar
- The width of the search bar seems off at this size and others

horatiorosa:
> I took a look at the mobile view and noticed some issues: image
> 
> * "About" is missing padding on the right.
> * The search bar should be in a second row
> * The width of the drop-down appears to be the same as the search bar
> * The width of the search bar seems off at this size and others

darn, good catch!
I'll take care of these issue.

app/components/HeaderBar/AddressSearch.tsx
```diff
@@ -0,0 +1,300 @@
...
+  const handleInputChange = (details: ComboboxInputValueChangeDetails) => {
+    filter(details.inputValue);
+  };
+
+  return (
+    <Combobox.Root
+      collection={collection as ListCollection<ComboboxCollectionItemProps>}
+      onInputValueChange={handleInputChange}
+      inputBehavior="autohighlight"
+      width="275px"
```
dhochbaum-dcp: This should be a max-width of 376px: https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=4305-43913&m=dev

horatiorosa: affirmative. I updated width on `base` and `md` breakpoints

app/components/HeaderBar/HeaderBar.tsx
dhochbaum-dcp, line 1: @TylerMatteo do you want this to be refactored so that all of the changes are behind the feature flag?  Or is it ok as-is, with only the address search behind the feature flag?

horatiorosa: It looks like I'll have to put the design changes behind a flag. Good call out on this. I'll update so that the all address ui changes behind the flag.

dhochbaum-dcp (dismissed):
Nice work, Horatio

dhochbaum-dcp (approved):

PR #413, opened 2026-03-23
399/filter projects and budget requests by borough

### LINKED ISSUE #399: Filter projects and budget requests by Borough
## Acceptance Criteria

- [ ] Add Boroughs option to new "map control" component ([designs](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=622-28087&m=dev))
- [ ] See #376 for detail on expected behavior

### Dependent
* Dependent on #376 
* Dependent on [Zoning API issue ](https://github.com/NYCPlanning/ae-zoning-api/issues/566) for filtering CBBRs by boroughs. Corresponding Capital Project Issue is already completed

pratishta:
Closes https://github.com/NYCPlanning/ae-cp-map/issues/399

- Implementation of filtering by borough
image

- Remove 'Filter by Location' section on left panel
- Change `districtType` and `districtId` to `boundaryType` and `boundaryId` respectively
- We use the same search param 'boroughId' for community district and borough filtering. To avoid collision, the selections are stored and updated in state and not in the URL


app/components/layers/useCapitalProjectsLayer.client.tsx
```diff
@@ -154,8 +154,14 @@ export function useCapitalProjectsLayer(opts: {
...
-      extensions: [maskExtension, dataFilterExtension],
-      maskId: `${districtId !== null ? "boundary-mvt" : ""}`,
+      extensions: [
+        new MaskExtension(),
+        new DataFilterExtension({
+          filterSize: 1,
+          categorySize: 2,
+        }),
+      ],
+      maskId: `${(boundaryId !== null && (boundaryType === "cd" || boundaryType === "ccd")) || (boroughId !== null && boundaryType === "borough") ? "boundary-mvt" : ""}`,
```
pratishta: I think this can be more concise

app/layouts/ResultsPanel.tsx
```diff
@@ -137,17 +140,20 @@ export async function loader({ request }: LoaderFunctionArgs) {
...
-      ...(districtId !== null && districtType === "ccd"
-        ? { cityCouncilDistrictId: districtId }
+      ...(boroughId !== null && boundaryType === "borough"
+        ? { boroughIds: [boroughId] }
         : {}),
       ...(managingAgency === null ? {} : { managingAgency }),
       ...(agencyBudget === null ? {} : { agencyBudget }),
       ...(commitmentsTotalMin === null ? {} : { commitmentsTotalMin }),
       ...(commitmentsTotalMax === null ? {} : { commitmentsTotalMax }),
-      ...(districtId !== null ? {} : { isMapped: true }),
```
pratishta, line 150: I'm still a little bit confused on why we check for this

TylerMatteo:
@pratishta taking a look at this now

app/layouts/ResultsPanel.tsx
TylerMatteo: The requirements when we first put in capital projects were that only mapped projects were shown. We put all projects into data flow and zoning API because it wasn't much difference in effort and we may need them eventually. This param filters to only mapped projects

TylerMatteo:
Documenting next steps from a conversation @pratishta and I had:

Pratishta will try to refactor `boundaryId` out the code in favor of dedicated query params for each boundary type (cd, ccd, and borough), while preserving the existing behavior where only the ID for the _active_ boundary type is kept in the path URL at any given time. This is a different approach than what David and I settled on in his previous PR but, with the complexity introduced by the addition of borough filtering, we feel like its a worthwhile change to try.

pratishta:
> Documenting next steps from a conversation @pratishta and I had:
> 
> Pratishta will try to refactor boundaryId out the code in favor of dedicated query params for each boundary type (cd, ccd, and borough), while preserving the existing behavior where only the ID for the active boundary type is kept in the path URL at any given time. This is a different approach than what David and I settled on in his previous PR but, with the complexity introduced by the addition of borough filtering, we feel like its a worthwhile change to try.

After taking a look, I believe these changes are out of scope for this ticket. The main reason is that this PR is already addressing a number of changes, and conditionally adding new query parameters behind a feature flag is, I believe, is not a worthwhile refactor. 

Adding new query parameters `cityCouncilDistrictId` and `communityDistrictId` changes the types we define in `utils/types.ts`. This requires updating the FilterMenu and the subsequent dropdown menus. I realize that we can create new types to only use in the new filtering UI, but doing so in this PR - which is about filtering by boroughs - adds another set of changes that really belong in a separate ticket. 

As is, I don't think there's much complexity in adding borough filtering other than a potentially confusing URL where `boroughId` can refer to a CD or a borough. But the presence of `boundaryType` should be clarifying in that case anyway. The issue of CDs having two query parameters (`boroughId` and `boundaryId`) whereas CCDs have one (`boundaryId`) has existed prior to adding borough filtering, and should be addressed separately. 

@TylerMatteo @dhochbaum-dcp

dhochbaum-dcp:
> > Documenting next steps from a conversation @pratishta and I had:
> > Pratishta will try to refactor boundaryId out the code in favor of dedicated query params for each boundary type (cd, ccd, and borough), while preserving the existing behavior where only the ID for the active boundary type is kept in the path URL at any given time. This is a different approach than what David and I settled on in his previous PR but, with the complexity introduced by the addition of borough filtering, we feel like its a worthwhile change to try.
> 
> After taking a look, I believe these changes are out of scope for this ticket. The main reason is that this PR is already addressing a number of changes, and conditionally adding new query parameters behind a feature flag is, I believe, is not a worthwhile refactor.
> 
> Adding new query parameters `cityCouncilDistrictId` and `communityDistrictId` changes the types we define in `utils/types.ts`. This requires updating the FilterMenu and the subsequent dropdown menus. I realize that we can create new types to only use in the new filtering UI, but doing so in this PR - which is about filtering by boroughs - adds another set of changes that really belong in a separate ticket.
> 
> As is, I don't think there's much complexity in adding borough filtering other than a potentially confusing URL where `boroughId` can refer to a CD or a borough. But the presence of `boundaryType` should be clarifying in that case anyway. The issue of CDs having two query parameters (`boroughId` and `boundaryId`) whereas CCDs have one (`boundaryId`) has existed prior to adding borough filtering, and should be addressed separately.
> 
> @TylerMatteo @dhochbaum-dcp

I agree that performing a switch of the existing query parameters is probably out of scope for this ticket, and we would be better served by incorporating that into part of the changes required to switch to allowing selection of multiple geographies.

I do think that this PR should be changed to use `boroughIds` as the parameter, as that will be what we will be using in the future, and it avoids any issues of collision with the existing `boroughId` parameter.

app/components/layers/useBoroughsLayer.client.tsx
```diff
@@ -0,0 +1,73 @@
...
+          });
+        } else {
+          updateSearchParams({
+            boroughIds: info.object.properties.id,
+          });
+        }
+      },
+      pointType: "text",
+      getText: ({ properties }: { properties: BoroughProperties }) =>
+        properties.id,
```
dhochbaum-dcp: I don't see anything about it in the designs, but I can't imagine we want the text to be the ids.  Can you add `abbr` and `title` to the BoroughProperties, update this to `properties.title`, and raise the text size to 25, and we can confirm with @Ntoal about what she wants exactly when she returns?

app/layouts/MapPage.tsx
```diff
@@ -249,11 +250,12 @@ export default function MapPage() {
...
+  const boundaryType = searchParams.get("boundaryType") as BoundaryType;
+  const boundaryId = searchParams.get("boundaryId") as BoundaryId;
   const boroughId = searchParams.get("boroughId") as BoroughId;
+  const boroughIds = searchParams.get("boroughIds") as BoroughId;
 
-  const getDefaultIndex = (type: DistrictType | null) => {
+  const getDefaultIndex = (type: BoundaryType | null) => {
     switch (type) {
       case "ccd":
         return 1;
```
dhochbaum-dcp, line 261: Loading a url where `boundaryType=borough` defaults to having Community Districts selected.  Adding `case "borough": return 2;` should fix the issue.

app/components/atlas.client.tsx
```diff
@@ -103,6 +110,9 @@ export function Atlas({
           capitalProjectBudgetedGeoJsonLayer,
           communityBoardBudgetRequestsLayer,
           communityBoardBudgetRequestGeoJsonLayer,
+          boroughsLayer,
+          boroughsOutlinesLayer,
+          boroughLayer,
         ]
```
dhochbaum-dcp, line 116: Adding these to the end of the array makes them the top layers, which causes the Capital Projects and CBBRs un-clickable.  Moving the borough layers to before the Capital Project and CBBR layers in the array will fix the issue.

TylerMatteo:
@pratishta I don't see any major issues aside from things @dhochbaum-dcp already called out.

app/components/layers/useCapitalProjectsLayer.client.tsx
pratishta: Going to leave it as is in favor of being as precise as possible for the boundary type conditions. Can revisit if filtering by multiple geos will complicate it further

dhochbaum-dcp (approved):

PR #416, opened 2026-03-26
Task/403 add search input to UI dh

horatiorosa:


horatiorosa:
no longer needed, thank you Hawk.

PR #421, opened 2026-03-30
402 & 404 Connect Geosearch, add Map Pin

### LINKED ISSUE #402: Add Address search input and map pin
### Summary
This will be the first of several issues to add address and radius filtering to CPP. The scope of this Issue will implement the feature in an incomplete state but will hide the changes behind a feature flag

#### Acceptance criteria

- [ ] Add new input to header bar using existing `Combobox` component in Streetscape
- [ ] User can search for addresses using Geosearch API and select a result. See [design and flow diagram in Figma](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=899-52143&m=dev) and AC in parent Story for details
- [ ] When user selects a result, a pin is added to the map as per designs
- [ ] All changes should be behind existing Facdb phase 1 feature flag

#### Out of scope
- [ ] The "Filter by Radius" button and popover are out of scope and will added following completion of this Issue.

### LINKED ISSUE #404: Add Map Pin to Selected Search Result
#### Acceptance criteria

- [ ] When a user selects an address from the address suggestions:
    - [x] the map must center and zoom to that address
    - [x] a map pin must be placed at the selected location
    - [ ] results must update immediately across all CPP datasets
- [ ] Clearing the address must:
    - [x] remove the map pin
    - [ ] return results to the prior unfiltered state
- [x] Selecting an address must clear any active geography filters, and selecting a geography must clear any active address filter.
- [x] All changes should be behind existing Facdb phase 2 feature flag

#### Out of scope
- [ ] Connecting the input to the Geosearch API is a separate issue
- [ ] The "Filter by Radius" button and popover are out of scope and will added following completion of this Issue.

### Dependent
* Dependent on #403

dhochbaum-dcp:
Closes #402 
Closes #404


app/address-search/AddressSearchResult.ts
TylerMatteo, line 1: Maybe we could define Zod schemas instead of types? That way we could actually validate the data at run time and [infer types](https://zod.dev/basics#inferring-types) from the schemas. Also, Idk if we necessary need to schema out the whole response. We could try defining a [loose object](https://zod.dev/api#zlooseobject) with only the sections we need.

app/utils/types.ts
```diff
@@ -47,14 +47,24 @@ export type PaginationQueryParams = {
   page?: number;
 };
 
+export type GeosearchQueryParams = {
```
TylerMatteo: Can we call this `AddressQueryParams` or something similar? Trying to keep the Geosearch references out of the app code where possible.

app/root.tsx
```diff
@@ -21,8 +21,30 @@ import { useEffect, useState } from "react";
 import { initializeMatomoTagManager } from "./utils/analytics";
 import { HeaderBar } from "./components/HeaderBar";
 import { INITIAL_VIEW_STATE } from "./components/atlas.client";
+import {
+  ComboboxInputValueChangeDetails,
+  ListCollection,
+  createListCollection,
+  useCombobox,
+} from "@ark-ui/react/combobox";
```
TylerMatteo: You're importing from ark here but its not listed as a dependency in `package.json`? It works because streetscape has it as a dep but you should always list things you're using directly as dependencies. Or, better yet, lets update streetscape to export these types and utilities so the app doesn't need the dependency.

app/root.tsx
```diff
@@ -63,17 +85,98 @@ function Document({
...
+      if (search !== null || radius !== null || pin !== null) {
+        updateSearchParams({
+          search: undefined,
+          radius: undefined,
+          pin: undefined,
+        });
+      }
+      setAddressSearchQuery(details.inputValue);
+    } else if (details.reason === "item-select") {
+      // do nothing, this is handled by handleselection
```
TylerMatteo: Do you need this `else if` block at all then?

app/components/HeaderBar/AddressSearch.tsx
```diff
@@ -276,7 +34,14 @@ export const AddressSearch = () => {
...
+        <Combobox.Content
+          backgroundColor={"white"}
+          display={
+            addressSearchQuery !== null && addressSearchQuery.length > 2
+              ? "block"
+              : "none"
+          }
+        >
           <Text
             style={{
```
TylerMatteo: This needs to be `sx`, not `style` for these styles to apply correctly. `style` just adds a regular html inline style. I won't be compiled to CSS and won't have access to the theme tokens. Most of these styles either aren't getting applied or are getting coerced to `px values`

Image

TylerMatteo (changes requested):

app/root.tsx
dhochbaum-dcp: PR to update this in streetscape is here: https://github.com/NYCPlanning/ae-streetscape/pull/142

app/address-search/AddressSearchResult.ts
dhochbaum-dcp: Loose objects were added in v4, we are currently on 3.25.56.  Do you want to upgrade to v4?  Stick with v3 and create a zod schema for the entire object?  Or something else?

app/utils/types.ts
dhochbaum-dcp: Yes, updated.

app/root.tsx
dhochbaum-dcp: No, it's removed.

app/components/HeaderBar/AddressSearch.tsx
dhochbaum-dcp: I moved them to be normal properties on the Text component.

app/address-search/AddressSearchResult.ts
TylerMatteo: Tbh I could live with either approach. I think Kubb also Zod dependencies and ideally we'd keep them all on the same major Zod version. I'd say try the upgrade approach and we can reevaluate if the upgrade itself is nontrivial.

dhochbaum-dcp: Ok, updated to zod4 and pushed

app/root.tsx
dhochbaum-dcp: Fixed.

app/components/HeaderBar/HeaderBar.tsx
```diff
@@ -91,7 +103,12 @@ export function HeaderBar({
           }}
           className="address-search"
         >
-          <AddressSearch />
+          <AddressSearch
+            combobox={combobox}
+            addressSearchQuery={addressSearchQuery}
+            addressSearchResults={addressSearchResults}
```
TylerMatteo: I noticed this is showing up even on pages that don't use `MapPage`. Maybe we can have `HeaderBar` take `children` then use `AddressSearch` inside it in `MapPage`?

app/address-search/axios/findAddresses.ts
```diff
@@ -0,0 +1,33 @@
...
+import type {
+  RequestConfig,
+  ResponseErrorConfig,
+} from "@kubb/plugin-client/clients/axios";
+
+/**
+ * @summary Find addresses
+ * {@link /addresses}
+ */
+export async function findAddresses(
```
TylerMatteo, line 18: Can we have this function run the `parse` method on the schemas to validate the response data? The value of using Zod, in addition to the type inference, is we can use the schemas to confirm that data coming back from the API of the expected shape.

TylerMatteo (changes requested):

app/components/HeaderBar/HeaderBar.tsx
dhochbaum-dcp: `HeaderBar` appears in `root` so if I understand what you're suggesting, I don't believe it's possible.  I pushed a fix which uses `useMatches` to conditionally show the `AddressSearch` on pages using the `MapPage` layout.

app/address-search/axios/findAddresses.ts
dhochbaum-dcp: Let me know if I've implemented that how you'd like - I'm not quite sure what you want to do if it fails.

app/components/HeaderBar/HeaderBar.tsx
```diff
@@ -21,6 +21,10 @@ export function HeaderBar({
   isLoading: boolean;
 }) {
   const [searchParams] = useSearchParams();
+  const matches = useMatches();
+  const isMapPage = matches
+    .map((match) => match.id)
+    .includes("layouts/MapPage");
```
horatiorosa, line 27: this is a neat and tidy solution    🎉

TylerMatteo: I'm cool with this for now but we should at least revisit it as something to refactor. A component shouldn't "care about" what route its rendered on.

app/components/HeaderBar/HeaderBar.tsx
TylerMatteo: I think what I had in mind would require moving the user of `<HeaderBar>` to the respective layout files, which as its own set of trade offs. See my [response](https://github.com/NYCPlanning/ae-cp-map/pull/421#discussion_r3041453416) to Horatio's comment.

app/address-search/axios/findAddresses.ts
TylerMatteo: So this isn't quite what I had in mind but looking into it more led me down a rabbit hole with @Ntoal that led to some [new error messaging](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=2227-81709&t=6Bgc3jb0ljgJQaKG-0) that wasn't in the original requirements. All of that to say we can go ahead with this and bundle the refactor and new error messages into a follow up Issue.

TylerMatteo (approved):

PR #433, opened 2026-04-10
424 Add Radius Filter Button and Popover

### LINKED ISSUE #424: Add Radius Filter Button and Popover
### Acceptance Criteria

- [ ] Add filter by radius button and panel
- [x] Applies radius filter around selected Address as per AC in parent issue
- [x] See designs [here](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=622-28086&t=IRJ0zJ0oVWYMi1Jy-0)

#### Out of scope
* Adding Address/radius filter to "Selected location" section in results panel

dhochbaum-dcp:
Added the menu to UI, implemented on map.

image

Closes #424


Ntoal:
Looks good! @dhochbaum-dcp

app/components/HeaderBar/RadiusDropdown.tsx
```diff
@@ -0,0 +1,175 @@
...
+          variant={"secondary"}
+          isDisabled={true}
+          maxH={10}
+          display={"flex"}
+          px={6}
+          py={3}
+          justifyContent={"center"}
+          alignItems={"center"}
+          border={"1px solid"}
+          borderColor={"gray.400 !important"}
```
horatiorosa: I think you can use the `sx={{ }}` prop to override base styles according to the [docs here](https://github.com/chakra-ui/chakra-ui/discussions/5306).

app/utils/constants.ts
TylerMatteo, line 1: Can this live in another existing file more related to its use case, maybe the one for the component? I worry starting a `constants` file is sometimes an antipattern as it becomes used for all sorts of things.

app/utils/utils.ts
```diff
@@ -10,6 +10,13 @@ import {
   QueryParams,
 } from "./types";
 
+export const formatDistance = (distance: number) => {
```
TylerMatteo, line 13: Unit test?

app/components/layers/useMapPinLayer.client.ts
```diff
@@ -1,25 +1,106 @@
...
+import { formatDistance } from "~/utils/utils";
 
-export function useMapPinLayer() {
+export type MapPinIconLayerProps = {
+  coordinates: number[];
+  radius: number;
+};
+
+export function useMapPinLayer({
+  addressSearchSliderValue,
```
TylerMatteo, line 31: Maybe we can make this argument name more general purpose? Could it be `radius` just like the `MapPinLayer` argument it gets passed to?

app/components/layers/useMapPinLayer.client.ts
```diff
@@ -1,25 +1,106 @@
...
+        id: "MapPinRadiusCircleScatterplotLayer",
+        data: [
+          {
+            coordinates: this.props.coordinates,
+          },
+        ],
+        visible:
+          this.props.coordinates !== undefined &&
+          this.props.radius !== undefined,
+        getFillColor: [79, 209, 197, 76],
```
TylerMatteo: Have you tried letting the _instance of this class_ set arguments like `getFillColor`, `getLineWidth`, `getBackgroundColor`, etc instead of hard coding them into the Class itself? What you're doing with `coordinates: this.props.coordinates` by forwarding props with `this.props` is the right idea, but ideally we'd do it for props like this as well.

TylerMatteo (dismissed):

app/components/layers/useMapPinLayer.client.ts
dhochbaum-dcp: Discussed on Teams - this is to differentiate the `radius` value in the URL parameter from the actual value of the slider, which is stored separately so that we can update the circle on the map as the user moves the slider.

app/components/HeaderBar/RadiusDropdown.tsx
dhochbaum-dcp: [sx](https://v2.chakra-ui.com/docs/styled-system/the-sx-prop) is only intended to be used if necessary - if we are able to set properties as above, doing so is best practice.

app/utils/utils.ts
dhochbaum-dcp: Added.

app/utils/constants.ts
dhochbaum-dcp: Sure, moved to `app/components/HeaderBar/AddressSearch.tsx`

app/components/layers/useMapPinLayer.client.ts
dhochbaum-dcp: Let me know if my implementation matches what you were thinking.

TylerMatteo: Yep lgtm!

TylerMatteo (approved):

PR #451, opened 2026-04-17
Task/425 add radius filter results to selected location v1

### LINKED ISSUE #425: Add Radius filter to Selected Location
### Acceptance Criteria

- [x] Add active Radius filter to "Selected Location" section of Results Panel ([designs](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=144-35727&t=p9YLw5kQoQMByeyJ-0))
- [x] User can clear the radius filter 

#### Dependent
* #424 
* #412

horatiorosa:
#### Tickets
Closes #425

### DIFF
--- a/app/components/HeaderBar/RadiusDropdown.tsx
+++ b/app/components/HeaderBar/RadiusDropdown.tsx
@@ -22,11 +22,13 @@ import { useUpdateSearchParams } from "~/utils/utils";
 import { ADDRESS_SEARCH_RADIUS } from "~/components/HeaderBar/AddressSearch";
 
 export const RadiusDropdown = ({
-  addressSearchSliderValue,
   setAddressSearchSliderValue,
+  clearRadiusFilter,
+  addressSearchSliderValue,
 }: {
   addressSearchSliderValue: number | undefined;
   setAddressSearchSliderValue: (v: number | undefined) => void | undefined;
+  clearRadiusFilter: () => void;
 }) => {
   const [searchParams, updateSearchParams] = useUpdateSearchParams();
   const pin = searchParams.get("pin");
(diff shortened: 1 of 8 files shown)


app/components/SelectedLocations.tsx
```diff
@@ -128,7 +153,7 @@ export function SelectedLocations() {
                 height={"8px"}
                 onClick={clearTag}
                 aria-label={"closeIcon"}
-                verticalAlign={"baseline"}
+                verticalAlign={"middle"}
```
horatiorosa, line 156: per @Ntoal , `middle` is the way to go.

app/root.tsx
```diff
@@ -204,6 +203,28 @@ export function Main() {
...
+    setAddressSearchSliderValue(undefined);
+    updateSearchParams({ radius: undefined });
+  };
+
+  const setDefaultRadiusFilter = () => {
+    if (addressSearchSliderValue === undefined) {
+      setAddressSearchSliderValue(ADDRESS_SEARCH_RADIUS.DEFAULT);
+      updateSearchParams({ radius: ADDRESS_SEARCH_RADIUS.DEFAULT });
+    }
+  };
```
dhochbaum-dcp: This function seems like overkill - it's only used once, in a component we're already passing `resetRadiusFilter`.  We can just pass `addressSearchSliderValue` to that component and do this check to see if it's defined in the component where it's used.

horatiorosa: Yeah, that is a good idea.  I will also move `resetRadiusFilter` back into `RadiusDropdown` for the same reason. It is only needed in one place.
I think I did that bc I thought I had to use that in `selectedLocations`. I didn't.
`¯\_(ツ)_/¯`

app/components/HeaderBar/RadiusDropdown.tsx
```diff
@@ -22,19 +22,33 @@ import { useUpdateSearchParams } from "~/utils/utils";
...
-      : addressSearchSliderValue;
+  const radiusValueInMiles = new Intl.NumberFormat("en", {
+    maximumFractionDigits: 2,
+  }).format(sliderValue / 5280);
+
+  const setDefaultRadiusFilter = () => {
+    if (addressSearchSliderValue === undefined) {
+      resetRadiusFilter();
+    }
+  };
```
dhochbaum-dcp: You don't need to define this function at all, you can just put the `if` statement directly in the `onOpen`.

dhochbaum-dcp (approved):
Great work!

PR #457, opened 2026-04-23
Update address search no results message and add error message.

### LINKED ISSUE #431: Add additional error handling for address search
### Acceptance Criteria

- [x] Update address search results with API error message and new no results copy as per [designs](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=2227-81709&t=xFunk4cBxnQ1DFdb-0)

TylerMatteo:
Closes #431 

### Summary

Updates the message shown to users when their address search returns no result
image

Adds a new error message shown to users when their address search fails for some reason. This can be tested by blocking requests to geosearch in dev tools. **Note you will still see "loading" for a couple seconds because Tanstack query defaults to retrying queries 3 times.** This error will also be shown if the network request to Geosearch is successful but the Zod `.parse` fails for some reason.
image

### DIFF
--- a/app/address-search/axios/findAddresses.ts
+++ b/app/address-search/axios/findAddresses.ts
@@ -18,11 +18,11 @@ import { findAddressesQueryResponseSchema } from "../zod";
 export async function findAddresses(
   text: FindAddressesQueryParams["text"],
   config: Partial<RequestConfig> & { client?: typeof fetch } = {},
-) {
+): Promise<FindAddressesQueryResponse> {
   const { client: request = fetch, ...requestConfig } = config;
 
   const res = await request<
-    FindAddressesQueryResponse,
+    unknown,
     ResponseErrorConfig<FindAddresses400 | FindAddresses500>,
     unknown
   >({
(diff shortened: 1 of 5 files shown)


app/address-search/axios/findAddresses.ts
TylerMatteo, line 1: @dhochbaum-dcp some of these changes are a refactor from your earlier PR. I changed the type on line 25 to `unknown` because its better practice IMO to assume we don't actually _know_ what the shape of that data will be. The return type of the whole function is still `Promise<FindAddressesQueryResponse>` because once we do the `.parse` with Zod, we know for sure that the data match that type.

I also removed the try catch because I'm not sure if its really necessary now that I have proper error handling. We want the error to "bubble up" and don't need to do much with it in a catch block in this file.

app/root.tsx
```diff
@@ -109,11 +109,17 @@ export function Main() {
...
+  const {
+    data: addressSearchResults,
+    isLoading,
+    error,
+  } = useQuery({
     queryKey: ["find-addresses", addressSearchQuery],
     queryFn: queryFunction,
   });
 
+  console.log({ error });
```
dhochbaum-dcp: Please delete, it's doing a console log every keystroke.

dhochbaum-dcp:
Looks good to me, please just remove the console log.

app/root.tsx
TylerMatteo: Whoops, meant to remove this.

dhochbaum-dcp (approved):

PR #472, opened 2026-05-05
468 Add ability to filter by multiple boroughs

dhochbaum-dcp:
- [x] User can select multiple boroughs at the same time for filtering
- [x] CPs and CBBRs are filtered by selected boroughs
- [x] If user switches boundary type and then switches back to boroughs, their selection is reapplied.
- [x] Functionality is hidden behind `VITE_FACDB_PHASE_3` feature flag

This also updates the "Selected Location" tags according to [the designs](https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=3363-145892&p=f&m=dev).
image

This uses a CompositeLayer which contains multiple GeoJSONLayers for the selected areas.  The class for the selected area has been written with the intent of being used by all geo selection types.

Closes #468


app/components/layers/SelectedGeosLayer.ts
```diff
@@ -0,0 +1,32 @@
+import { env } from "~/utils/env";
+import { GeoJsonLayer } from "@deck.gl/layers";
+import { CompositeLayer, LayersList } from "@deck.gl/core";
+
+const { zoningApiUrl } = env;
```
TylerMatteo: Can we decouple this from class from `env`? Maybe the hook can take care of this part and pass it in as part of `getDataUrlPath`.

app/components/layers/SelectedGeosLayer.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export type SelectedGeosLayerProps = {
+  geoIds: null | string[];
```
TylerMatteo: Lets go with just `ids`.

app/components/layers/useBoroughLayer.client.tsx
TylerMatteo, line 1: How feasible would it be to keep this file and use it behind the feature flag to preserve the existing functionality?

TylerMatteo (changes requested):

app/components/layers/SelectedGeosLayer.ts
dhochbaum-dcp: I can make that change if you want, but is there a benefit?  It just seems cleaner to do it once in this class, rather than 3 times in each of the functions.

app/components/layers/useBoroughLayer.client.tsx
dhochbaum-dcp: In terms of the code, it leaves multiple extra files in the repo, and has us using extra layers, using more complicated if statements, and having more merge conflicts until after the phase 3 release.  Getting rid of it and hiding the functionality behind a single if statement is much cleaner.  If you're worried that the change could introduce issues into production before the phase 3 launch, it's likely that those issues would exist after the phase 3 launch; in that case, we're better off knowing sooner rather than later.

Please no.

app/components/layers/SelectedGeosLayer.ts
TylerMatteo: I appreciate not wanting to repeat code but in this case, I think the benefit of decoupling the layer from `env` is worth the trade off.

app/components/layers/useBoroughLayer.client.tsx
TylerMatteo: I might be cool with making an exception to feature flagging here, but we know we have least one release coming before this functionality is supposed to be released. We're supposed to release "phase 2" (aka address search) as soon as next week. This "multi select" functionality isn't supposed to be a part of that, so merging this puts us in an awkward situation.

dhochbaum-dcp: My mistake - I think I misinterpreted your question.  There is a feature flag in the `onClick` of `useBoroughsLayer`.  Users will not be able to select multiple boroughs by clicking on the map until the phase 3 flag is enabled.

TylerMatteo (approved):

app/components/layers/useBoroughLayer.client.tsx
TylerMatteo: Ah my bad, I totally missed that.

PR #487, opened 2026-05-29
485 Revise CBBR filtering by agency response checkbox behavior

dhochbaum-dcp:
This adds Zustand to use as the Store, to allow for the switch to having checkboxes checked by default.

Closes #485


horatiorosa (approved):
I reviewed / read the zustand docs (and while I am no zustand expert) everything in the store you created seems to follow the example. 

Tested in browser and the agency checkboxes are all check by default.

pratishta (approved):

PR #512, opened 2026-06-26
500 CPP/FacDB - Add Facilities to Map View

dhochbaum-dcp:
- Add the toggle to filter menu
- Add dots to the map
- Geographic filters should all work

Closes #500


app/components/layers/useFacilitiesLayer.client.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const bufferParam = searchParams.get("radius");
+  const buffer = bufferParam === null ? -1 : parseInt(bufferParam);
+  const pin = searchParams.get("pin");
+  const [lon, lat] =
+    pin === null
+      ? [undefined, undefined]
+      : pin.split(",").map((d) => parseFloat(d));
+
+  return new MVTLayer<FacilityProperties, MaskExtensionProps>({
+    id: "FacilitiesDistricts",
```
pratishta: Let's call this `facilities`

app/components/layers/useFacilitiesLayer.client.tsx
```diff
@@ -0,0 +1,132 @@
...
+  visible,
+}: {
+  viewState: MapViewState;
+  setViewState: (newViewState: MapViewState) => void;
+  visible: boolean;
+}) {
+  const { facilityId } = useParams();
+  const [searchParams] = useSearchParams();
+  const navigate = useNavigate();
+  const [hoveredFacility, setHoveredFacility] = useState<string | undefined>();
```
pratishta: Can we change this to `<string | null>` instead of allowing returning undefined?

app/components/layers/useFacilitiesLayer.client.tsx
```diff
@@ -0,0 +1,132 @@
...
+    iconSizeUnits: "meters",
+    iconSizeMaxPixels: isMobile ? 59 : 29,
+    getIcon: ({ properties }: { properties: FacilityProperties }) => ({
+      url: `data:image/png;base64,${FACILITY_PNG_BY_CATEGORY[properties.id === hoveredFacility ? "HOVER" : properties.id === facilityId ? "SELECTED" : "DEFAULT"].get(properties.categoryId)}`,
+      id: `facility-${properties.categoryId}${properties.id === facilityId && "-selected"}${properties.id === hoveredFacility && "-hovered"}`,
+      width: 132,
+      height: 132,
+      mask: false,
+    }),
+    onClick: (info) => {
```
pratishta, line 95: If this PR isn't meant to implement the onClick functionality, let's comment this out or remove it. Or handle the error

app/components/atlas.client.tsx
```diff
@@ -123,6 +132,7 @@ export function Atlas({
           communityBoardBudgetRequestsLayer,
           communityBoardBudgetRequestGeoJsonLayer,
           mapPinLayer,
+          facilitiesLayer,
```
pratishta: The designs show facilities behind cbbrs but in front of capital projects. If we moved `facilitiesLayer` it would follow the designs and also make it a bit easier on the eyes. 
```
          capitalProjectsLayer,
          capitalProjectBudgetedGeoJsonLayer,
          facilitiesLayer,
          communityBoardBudgetRequestsLayer,
          communityBoardBudgetRequestGeoJsonLayer,
          mapPinLayer,
```

app/components/SearchByCbbrMenu.tsx
```diff
@@ -89,104 +89,100 @@ export const SearchByCbbrMenu = ({
...
+      >
+        <Heading
+          flex="1"
+          textAlign="left"
+          fontSize="xs"
+          fontWeight="bold"
+          lineHeight="32px"
+          paddingBottom={0}
+        >
+          {`Filters (${appliedFilters.reduce((acc, curr) => acc + curr, 0)})`}
```
pratishta, line 106: Is there a reason we wanted to remove the 'hide/show' filters? Do we want to do the same for the hide/show legend in for cbbr filter menu?

app/components/layers/useFacilitiesLayer.client.tsx
```diff
@@ -0,0 +1,132 @@
+import { MVTLayer } from "@deck.gl/geo-layers";
+import { useState } from "react";
+import { env } from "~/utils/env";
+import { useNavigate, useParams, useSearchParams } from "react-router";
+import { FlyToInterpolator } from "@deck.gl/core";
+import { MaskExtension } from "@deck.gl/extensions";
+import type { MaskExtensionProps } from "@deck.gl/extensions";
+import type { MapViewState } from "@deck.gl/core";
+import { useMediaQuery } from "@nycplanning/streetscape";
+import { FACILITY_PNG_BY_CATEGORY } from "~/utils/facilities-icons";
```
pratishta, line 10: Should we have all the icons in one place and put this in public with cbbr icons?

pratishta (changes requested):
Looking awesome

app/components/layers/useFacilitiesLayer.client.tsx
dhochbaum-dcp: If we put it in public, it's actually a file we're hosting, which is not the intention.  We just want it to be included in the build.

app/components/SearchByCbbrMenu.tsx
dhochbaum-dcp: That was a requested change in the design docs: https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=5671-57758&m=dev

"Show Legend" was unchanged.

app/components/atlas.client.tsx
dhochbaum-dcp: Good catch

pratishta (approved):

PR #513, opened 2026-06-29
383 CPP/FacDB - Create FacDB record Detail page

### LINKED ISSUE #383: CPP/FacDB - Create FacDB record Detail page
As a CPP user, I want to open a Facilities record detail view. so that I can see key facility information and understand where the data came from.
Acceptance Criteria
1.	When a user selects a Facility from the map or Facilities results list, the system must open a Facilities Details view in the right-side panel.
2.	The Facilities Details view must display the following fields with data (when available):
a.	Facility / Program Name = FACNAME
b.	Address = ADDRESS
c.	BBL (Tax Lot ID) = BBL
d.	BIN (Building ID) = BIN
e.	Operated by = OPNAME
f.	Overseen by = OVERAGENCY
3.	The Facilities Details view must include a Source Data section that displays:
a.	Source = DATASOURCE
b.	Updated: date (xx/xx/xxxx) = Last Updated
c.	Display “Notice an issue with this data? Help us keep this site up to date; E-mail us (link to caps@planning.nyc.gov)
4.	The Facilities Details view must support navigation back to the Facilities results list.

 
Image

Prototype is from https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=90-22649&p=f&t=AxeoNyn6tIW1rlCC-0

dhochbaum-dcp:
Closes #383

### DIFF
--- a/app/components/FacilityDetail/index.tsx
+++ b/app/components/FacilityDetail/index.tsx
@@ -1,0 +1,477 @@
+import { Agency, Facility } from "~/gen";
+import {
+  Accordion,
+  AccordionButton,
+  AccordionIcon,
+  AccordionItem,
+  AccordionPanel,
+  ChevronLeftIcon,
+  EmailInvertedIcon,
+  ExternalLinkIcon,
+  Flex,
+  Heading,
+  HStack,
+  IconButton,
+  Link,
+  Tag,
+  Text,
+  VStack,
+  Button,
+} from "@nycplanning/streetscape";
+import { useNavigate, useSearchParams } from "react-router";
+import { format } from "date-fns";
+import { env } from "~/utils/env";
+
+const { stateOfGoodRepair } = env;
+
+export type FacilityDetailProps = {
+  facility: Facility;
+  facilityAgencies: Agency[];
+  categoryPath: string;
+  onNavigationClick: () => void;
+};
+
+const FACILITY_CATEGORY_ID_COLORS = new Map<number, string>([
+  [1, "#F0CB32"],
+  [2, "#58AE57"],
+  [3, "#EB9028"],
+  [4, "#86E3F3"],
+  [5, "#4977FA"],
+  [6, "#B66AC5"],
+  [7, "#8E8EA9"],
+]);
+
+export function FacilityDetail({
+  facility,
+  facilityAgencies,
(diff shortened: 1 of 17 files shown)


app/components/FacilityDetail/index.tsx
```diff
@@ -0,0 +1,480 @@
...
+          </Text>
+          <Text paddingLeft={3}>
+            <span style={{ fontWeight: "bold" }}>Type</span>:{" "}
+            {facility.facilityOperatorType === null
+              ? "Not specified"
+              : facility.facilityOperatorType}
+          </Text>
+        </HStack>
+
+        {stateOfGoodRepair === "ON" && facility.sgrLtr !== undefined && (
```
pratishta:
```suggestion
        {stateOfGoodRepair === "ON" && facility.sgrLtr !== null && (
```

sgrLtr (and the others) should never be undefined, only null or the enum values. Otherwise, I get a default grade for all non-graded facilities 

Image

app/components/FacilityDetail/index.tsx
```diff
@@ -0,0 +1,480 @@
...
+]);
+
+export function FacilityDetail({
+  facility,
+  facilityAgencies,
+  categoryPath,
+  onNavigationClick,
+}: FacilityDetailProps) {
+  // Parse date and remove leading 0's
+  const [year, month, day] = facility.dataSource.retrieveDate
```
pratishta:
```suggestion
    const date = format(facility.dataSource.retrieveDate, "M/dd/yyyy");  
```

`retrieveDate` is stored as a Date type in the database so we don't want to just parse and use it as a string if we can help it. We use the package date-fns in other parts of this repo so we can just import it and render `date`.

pratishta (changes requested):

app/components/FacilityDetail/index.tsx
dhochbaum-dcp: Good call, made the update.

app/components/FacilityDetail/index.tsx
dhochbaum-dcp: Good catch, made the update.

pratishta (approved):

PR #517, opened 2026-07-01
501 CPP/FacDB - Add Jurisdiction Filter

dhochbaum-dcp:
Closes #501 

Also makes list filter for facilityTypes, and cleans up a few unused variables.

### DIFF
--- a/app/components/SearchByFacilityMenu.tsx
+++ b/app/components/SearchByFacilityMenu.tsx
@@ -8,12 +8,17 @@ import {
 import { useDismissWelcomeAndUpdateSearchParams } from "../utils/utils";
 import {
   FacilityOversightAgency,
+  FacilityJurisdiction,
+  FacilityJurisdictions,
   FacilityType,
   FacilityTypes,
 } from "../utils/types";
 import { ClearFilterBtn } from "./ClearFilter";
-import { FacilityTypeCheckbox } from "./CheckboxControl";
 import { OversightAgencyDropdown } from "./DropdownControl";
+import {
+  FacilityJurisdictionCheckbox,
+  FacilityTypeCheckbox,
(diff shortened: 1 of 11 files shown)


app/components/SearchByFacilityMenu.tsx
```diff
@@ -26,14 +34,23 @@ export const SearchByFacilityMenu = ({
   const facilityTypeIds = useStore((state) => state.facilityTypeCheckboxes)
     .filter((ft) => ft.checked === true)
     .map((ft) => ft.id);
-  const updateFacilityTypeCheckboxById = useStore(
-    (state) => state.updateFacilityTypeCheckboxById,
-  );
+  const facilityJurisdictionIds = useStore(
+    (state) => state.facilityJurisdictionCheckboxes,
+  )
+    .filter((fJ) => fJ.checked === true)
```
pratishta: nit: if we have `ft` we should have `fj` instead of `fJ`

app/store/facility-jurisdiction.ts
```diff
@@ -0,0 +1,54 @@
+import { StateCreator } from "zustand";
+import { OversightLevelCategory } from "~/gen";
+import { FacilityJurisdiction } from "~/utils/types";
+
+interface FacilityJurisdictionCheckboxProps {
+  id: "City" | "State" | "Federal" | "County" | "Not specified";
```
pratishta: I'd rename this from `id` to `jurisdiction` or `label` or `name`. Same in facility-type.ts. I see them as different from the cbbr checkboxes because we actually do assign an id to each agency response category.

pratishta (changes requested):

pratishta (approved):

PR #520, opened 2026-07-07
Also at location link fix

dhochbaum-dcp:
This makes it so that the "Also at this location" links are actual links - previously you were unable to open them in a new window.

### DIFF
--- a/app/components/FacilityDetail/index.tsx
+++ b/app/components/FacilityDetail/index.tsx
@@ -454,11 +454,13 @@ export function FacilityDetail({
                   )}
                   borderRadius={"2px"}
                 />
-                <Text
+                <Link
                   color={"primary.600"}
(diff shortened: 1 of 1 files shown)


pratishta (changes requested):
Can we also use this PR to fix the colors for facility categories? The colors in the results panel and the also at this location section are wrong. 
**Example for results panel:** 
This is light blue which should be libraries 
Image
But this is core infrastructure 
Image

The fix is to just update `FACILITY_CATEGORY_ID_COLORS` with the right colors. I'd also suggest putting this constant in `app/utils/types.ts` so that it's defined in one place and we'd only have to update it once if we need to again

pratishta (approved):
Category colors addressed in https://github.com/NYCPlanning/ae-cp-map/pull/523

PR #522, opened 2026-07-07
494 CPP/FacDB/SGR - Update CPP Facilities About Page with State of Good Repair information

dhochbaum-dcp:
Updates the content of the About page for FacDB.

Closes #494

### DIFF
--- a/app/routes/about.tsx
+++ b/app/routes/about.tsx
@@ -15,6 +15,9 @@ import {
   UnorderedList,
   VStack,
 } from "@nycplanning/streetscape";
+import { env } from "~/utils/env";
+
+const { stateOfGoodRepair } = env;
 
 export default function About() {
   return (
@@ -233,8 +236,8 @@ export default function About() {
               </ListItem>
               <ListItem>
                 Not all Capital Projects in the Capital Commitment plan have a
-                spatial location – on average less than 40% of Capital Projects
-                in the plan are mapped in the web map on display. To review, all
+                spatial location - on average less than 40% of Capital Projects
+                in the plan are mapped in the web map on display. To review all
                 Capital Projects in CPDB, please refer to the tabular version
                 available at{" "}
                 <Link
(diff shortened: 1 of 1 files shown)


app/routes/about.tsx
```diff
@@ -556,6 +559,331 @@ export default function About() {
...
+              <ExternalLinkIcon marginLeft={0.5} />
+            </Link>
+            . Each score reflects the physical condition of a facility at the
+            time of its most recent assessment. Scores convey a general sense of
+            asset condition and should not be interpreted as an estimate of
+            capital investment needs or a prioritization of repairs.
+          </Text>
+          <Text pb={6}>
+            Not all facilities in FacDB have an associated SGR score; coverage
+            varies depending on whether a facility has been assessed and meet
```
horatiorosa: I know this is from the approved copy but the syntax is wrong, it should be `meets` to agree with the singular subject `facility`


`The difference between meet and meets comes down to basic subject-verb agreement. You use meet with plural subjects and the pronouns I, you, we, and they. You use meets exclusively with singular subjects—specifically, he, she, it, or singular nouns (like "John" or "the team")—in the simple present tense.`

app/routes/about.tsx
```diff
@@ -556,6 +559,331 @@ export default function About() {
...
+            >
+              Section 204
+              <ExternalLinkIcon marginLeft={0.5} />
+            </Link>{" "}
+            of the NYC Charter, the Facilities Database now includes additional
+            column information related to State of Good Repair (SGR) data for a
+            select number of New York City&apos;s owned and leased facilities.
+            SGR scores are gathered and scored using a standardized methodology
+            developed by the Office of Management and Budget (OMB), Department
+            of Design and Construction (DDC) and DCP&apos;s and sourced from
```
horatiorosa: I don't think DCP needs the `'s`, DCP here is just reference to thje development of the methodology along with the other city agencies mentions.

Image

app/routes/about.tsx
```diff
@@ -233,7 +236,7 @@ export default function About() {
               </ListItem>
               <ListItem>
                 Not all Capital Projects in the Capital Commitment plan have a
-                spatial location – on average less than 40% of Capital Projects
+                spatial location - on average less than 40% of Capital Projects
                 in the plan are mapped in the web map on display. To review, all
```
horatiorosa: probably unrelated to FacDB / SGR, but the comma here is wrong. It should read `To review all Capital Projects in CPDB,`

horatiorosa (changes requested):
very small nits but overall, ready to ship

horatiorosa (approved):
🚢

PR #523, opened 2026-07-08
504 CPP/FacDB - Add Facility Category Filter

dhochbaum-dcp:
Closes #504


pratishta (approved):
Looks good, two notes:

1. We should use deck.gl filter extension for filtering on the map instead of changing the icon size
2. We might want to separate out the updating query param function from the store specific file 

Approving so we can move forward with MVP for demo

PR #532, opened 2026-07-14
525 Facilities improvement - Use data filter extension to filter facilities

dhochbaum-dcp:
Switch from filtering facilities by icon size to using DataFilterExtension.

Closes #525


horatiorosa (approved):
LGTM

PR #542, opened 2026-07-31
538 CPP/FacDB - Update Export Modal

### LINKED ISSUE #538: CPP/FacDB - Update Export Modal
Based on designs: https://www.figma.com/design/syxdP7UMhpsgBnlD6hjvst/FacDB-Design?node-id=6389-69628&m=dev

dhochbaum-dcp:
Closes #538


app/utils/types.ts
dhochbaum-dcp, line 1: These changes re-organize some things to make it clearer what dataset each parameter is associated with.

app/root.tsx
dhochbaum-dcp, line 1: This is a bugfix, making a selection from the search drop-down had been clearing whatever previously selected geos were visible on the map, but was not making the changes to the URL

app/routes.ts
dhochbaum-dcp, line 1: The id is required in order for useRouteLoaderData to work

app/store/export-data-modal.ts
dhochbaum-dcp, line 1: We need to use the store to hold a lot of data with regard to the Export Data Modal because the "clear" buttons are designed to only change the settings of the exported csv, and not change any of the settings on the map or in the url.

app/store/export-data-modal.ts
```diff
@@ -0,0 +1,500 @@
...
+        lons === null ? undefined : lons.split(",").map((x) => parseFloat(x)),
+      lats:
+        lats === null ? undefined : lats.split(",").map((x) => parseFloat(x)),
+      geometry: buffer === null ? undefined : "Point",
+      limit: 1,
+      offset: 0,
+    };
+  },
+  async fetchUpdatedCapitalProjectsTotals() {
+    const { getUpdatedTotalsQueryParams, totalRecords } = get();
```
dhochbaum-dcp, line 341: Upon clicking any of the clear buttons, we need to update the number of records shown at the top of the modal.  The clear buttons for the three datasets each only require that we update the number of records shown when the user selects that dataset with the ratio button.  Clearing the selected geo requires updating stored value for each of the three datasets.

app/components/ExportDataModal/ExportDataModal.tsx
```diff
@@ -0,0 +1,509 @@
...
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+  Button,
+  Radio,
+  RadioGroup,
+  Text,
+  Switch,
```
pratishta: Are we using this?

app/components/ExportDataModal/ExportDataModal.tsx
```diff
@@ -0,0 +1,509 @@
...
+  Stack,
+  StreetscapeProvider,
+  HStack,
+} from "@nycplanning/streetscape";
+import { useEffect, useState } from "react";
+import { LinkBtn } from "../LinkBtn";
+import { useSearchParams, useLocation, useRouteLoaderData } from "react-router";
+import { env } from "~/utils/env";
+import { useStore } from "~/store";
+import { findFacilities } from "~/gen";
```
pratishta: Are we using this?

app/components/ExportDataModal/ExportDataModal.tsx
```diff
@@ -0,0 +1,509 @@
...
+                          </HStack>
+                        )}
+                      </Stack>
+                    </Stack>
+                  </RadioGroup>
+                </FormControl>
+              </Box>
+            </ModalBody>
+            <ModalFooter>
+              <LinkBtn
```
pratishta, line 511: Clicking this errors out on a new tab, should I be able to? Can we disable until we actually link to the download endpoint?

app/components/ExportDataModal/ExportDataModal.tsx
```diff
@@ -0,0 +1,509 @@
...
+      facilityOversightAgency: facilityOversightAgencyFromQueryParams,
+      facilityJurisdictions: facilityJurisdictionsFromQueryParams,
+      facilityCategoryIds: facilityCategoryIdsFromQueryParams,
+      facilityGroupIds: facilityGroupIdsFromQueryParams,
+      facilitySubgroupIds: facilitySubgroupIdsFromQueryParams,
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isOpen]);
+
+  const downloadLink = `${cpdbDataUrl}/${getExportDataModalDownloadQuery()}`;
```
pratishta: We should replace `cpdbDataUrl` with `zoningApiUrl` because right now, I get an error.

pratishta (changes requested):
Can't figure out why but I can't get a CSV that's facilities filtered by facility type (after replacing `cpdbDataUrl` with `zoningApiUrl`)

app/components/ExportDataModal/ExportDataModal.tsx
dhochbaum-dcp: Nope, removed.

app/components/ExportDataModal/ExportDataModal.tsx
dhochbaum-dcp: Nope, removed.

app/components/ExportDataModal/ExportDataModal.tsx
dhochbaum-dcp: Good call, that ENV variable is no longer needed, I've pushed a commit that removes it

app/components/ExportDataModal/ExportDataModal.tsx
dhochbaum-dcp: That shouldn't be the case if you're running against the up-to-date api running locally - can you please double check?

app/components/ExportDataModal/ExportDataModal.tsx
```diff
@@ -0,0 +1,508 @@
...
+                        )}
+                      </Stack>
+                    </Stack>
+                  </RadioGroup>
+                </FormControl>
+              </Box>
+            </ModalBody>
+            <ModalFooter>
+              <LinkBtn
+                isExternal
```
pratishta, line 512: Is there a benefit to this? Downloading the file in the same tab seems like a more seamless UX

pratishta (changes requested):
1. Are we implementing the toast? 
Image

2. What is the behavior supposed to be when you export 0 results? I forced 0 results in facilities by unselecting all facility categories and I get an export as if I didn't make that filter. 
Image

3. We should run kubb:generate and include the new csv generated types and files 

4. Why do we build and go to `downloadLink` directly with `getExportDataModalDownloadQuery` instead of calling and handling the individual CSV endpoints? 

5. Let's squash to 1 commit

app/components/ExportDataModal/ExportDataModal.tsx
pratishta: This comment was before I realized it was using the wrong domain instead of zoningapiurl

app/components/ExportDataModal/ExportDataModal.tsx
dhochbaum-dcp: If there's an error, you're taken away from the page.

app/components/ExportDataModal/ExportDataModalLinkButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export function ExportDataModalLinkButton({
+  downloadLink,
+  zeroTotalRecords,
+}: ExportDataModalLinkButtonProps) {
+  const toast = useToast();
+
+  const onClick = () => {
+    let filename = `CPP_Export_${String(new Date().getDate()).padStart(2, "0")}_${["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"][new Date().getMonth()]}_${new Date().getFullYear()}.csv`;
```
pratishta:
```suggestion
    let filename = `CPP_Export_${format(new Date(), "dd_LLL_yyyy")}.csv`;
```

app/components/ExportDataModal/ExportDataModalLinkButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    const downloadPromise = fetch(downloadLink)
+      .then((response) => {
+        if (!response.ok) throw new Error("Network response failed");
+
+        const disposition = response.headers.get("Content-Disposition");
```
pratishta: Not sure what this does, can you explain

pratishta (changes requested):
Some questions

app/components/ExportDataModal/ExportDataModalLinkButton.tsx
dhochbaum-dcp: That code doesn't work, there's no built-in "format" function.

app/components/ExportDataModal/ExportDataModalLinkButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+              href="mailto:CAPS@planning.nyc.gov"
+              textDecoration={"underline"}
+              color={"primary.600"}
+            >
+              E-mail us
+            </Link>
+            .
+          </Text>
+        ),
+        duration: null,
```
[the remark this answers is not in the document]
dhochbaum-dcp: Confirmed with @Ntoal, toasts should automatically disappear only upon success.

app/components/ExportDataModal/ExportDataModalLinkButton.tsx
dhochbaum-dcp: The filename above is a default to use only if the API returns a file without a filename specified in the headers.  We use a regex to search the Content-Disposition in the headers for a filename, and if we find one, we use that instead of the default.

app/components/ExportDataModal/ExportDataModalLinkButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+        position: "bottom-right",
+        isClosable: true,
+        title: "Success!",
+        description: "Export saved to your downloads folder.",
+        duration: 10000,
+      },
+      error: {
+        position: "bottom-right",
+        isClosable: true,
+        title: "Error. Please try again.",
```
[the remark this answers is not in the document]
dhochbaum-dcp: Good catch, I didn't realize the designs were updated between when I added them to Streetscape and now.  I've updated this.

app/components/ExportDataModal/ExportDataModalLinkButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+            </Link>
+            .
+          </Text>
+        ),
+        duration: null,
+      },
+      loading: {
+        position: "bottom-right",
+        isClosable: true,
+        title: "Export in progress...",
```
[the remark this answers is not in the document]
dhochbaum-dcp: Good catch, I didn't realize the designs were updated between when I added them to Streetscape and now.  I've updated this.

app/components/ExportDataModal/ExportDataModalLinkButton.tsx
pratishta: You need to import date-fns. 
`import { format } from "date-fns";` We use it in other places in the codebase.

pratishta (approved):

